PR #3, opened 2023-03-14
chore(*): add a CI pipeline to check pull requests

caohe:
#### What type of PR is this?
Chore

The original pipeline is used for building and pushing images during the release process. Therefore, we have added a new CI pipeline to check pull requests, which includes linting, building, and unit testing.


.github/workflows/release.yml
```diff
@@ -9,7 +9,6 @@ on:
   push:
     branches: [ 'master', 'main', 'release-*' ]
     # Publish semver tags as releases.
-  pull_request: {}
```
pendoragon, line 12: maybe these lines can be removed as well?
https://github.com/kubewharf/katalyst-core/pull/3/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R78-R82
and we can change this line to `push: true`
https://github.com/kubewharf/katalyst-core/pull/3/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R93

caohe: Yes, I think these lines are no longer necessary, and I have modified them.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    types: [ opened, edited, reopened, synchronize ]
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-18.04
```
pendoragon, line 13: seems ubuntu-18.04 will be deprecated soon https://github.com/actions/runner-images/issues/6002 
use ubuntu-latest instead?

caohe: "ubuntu-latest" refers to version 22.04, in which cgroup v2 is enabled by default and causes some unit tests to fail. Therefore, I switched the system version to the older 18.04. However, I also tried version 20.04 and it worked as well, which I consider to be a viable option.

pendoragon (approved):

PR #24, opened 2023-04-18
synchronize getting checkpoint

sun-yuliang:
#### What type of PR is this?
Enhancements

Synchronize getting checkpoint from cpu server to avoid race


waynepeking348 (approved):

pkg/agent/sysadvisor/plugin/qosaware/qos_aware.go
```diff
@@ -105,6 +105,7 @@ func (qap *QoSAwarePlugin) Run(ctx context.Context) {
 	go wait.UntilWithContext(ctx, qap.periodicWork, qap.period)
 
 	go qap.qrmServer.Run(ctx)
+	go qap.resourceAdvisor.Run(ctx)
```
cheney-lin, line 108: It seems that unnecessary to create new goroutine.

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -121,13 +117,52 @@ func NewCPUResourceAdvisor(conf *config.Configuration, extraConf interface{}, me
...
+	}
+
+	totalHeadroom := resource.NewQuantity(0, resource.DecimalSI)
+	for _, r := range cra.regionMap {
+		headroom, err := r.GetHeadroom()
+		if err != nil {
+			return headroom, err
+		}
+		// FIXME: is it reasonable to simply add headroom together?
+		totalHeadroom.Add(headroom)
```
cheney-lin, line 157: Empty numa which not contain any dedicated_cores and share_cores is not considered here, I will fix it in the future PRs.

luomingmeng (approved):

pkg/agent/sysadvisor/plugin/qosaware/qos_aware.go
sun-yuliang: Run works in a blocking way, waiting context to be done. So a new goroutine is needed

cheney-lin: resourceAdvisor.Run will also create new goroutine

PR #29, opened 2023-04-20
Dev/bugfix and refine

cheney-lin:
#### What type of PR is this?

Bug fixes


waynepeking348:
maybe we should refine the commit message and pr message more meaningful, for instance, which bug this pr aims to fix, rather than make every commit as `bug fix`

waynepeking348:
also, if you fix several different bugs, we suggest you to create different pull requests rather than mess up all up in a single one

cheney-lin:
this PR is not formal actually, former PRs have include some of these bug fixes.

PR #35, opened 2023-04-20
fix(sysadvisor): not clear binding numas any more

cheney-lin:
binding numas of dedicated_cores is fixed, and binding numas of share_cores is updated when updateNonExclusiveNumas, so we have no need to clear them.

#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_base.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_base.go
@@ -132,7 +132,6 @@ func (r *QoSRegionBase) Clear() {
 	r.Lock()
 	defer r.Unlock()
 
-	r.bindingNumas = machine.NewCPUSet()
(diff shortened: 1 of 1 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_base.go
```diff
@@ -132,7 +132,6 @@ func (r *QoSRegionBase) Clear() {
 	r.Lock()
 	defer r.Unlock()
 
-	r.bindingNumas = machine.NewCPUSet()
```
waynepeking348, line 135: how QoSRegionBase matches with `not clear binding numas for dedicated cores`

cheney-lin: fixed

waynepeking348 (approved):

sun-yuliang (approved):

PR #38, opened 2023-04-20
refactor(sysadvisor): emit metrics even when reclaim disabled

cheney-lin:
#### What type of PR is this?

Enhancements


waynepeking348:
why we still need to emit reclaimed resources when reclaim is disabled?

cheney-lin:
> 
Make the metrics continuous. Image that if the metrics between time1 to time2 is missed, the front-end grafana will connect the two point directly, giving the impression that the headrooms in this time span is non-zero.

sun-yuliang (approved):

luomingmeng (approved):

PR #42, opened 2023-04-21
fix(sysadvisor): maximize share pool size when disable reclaim

sun-yuliang:
#### What type of PR is this?
Bug fixes

Maximize share pool size when enable reclaim is false instead of still returning normal reclaim pool cpuset

It's still possible to trigger cpu plugin fallback logic, with potential bugs


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_share.go
```diff
@@ -109,6 +109,12 @@ func (r *QoSRegionShare) GetProvision() (types.ControlKnob, error) {
 			klog.Errorf("GetControlKnobAdjusted by policy %v err %v", internal.name, err)
 			continue
 		}
+		if !r.EnableReclaim {
```
cheney-lin, line 112: It's better to move this section before policy.GetControlKnobAdjusted

PR #44, opened 2023-04-24
add util to read file into string slices

waynepeking348:
#### What type of PR is this?
enhancements

sometimes we need to parse data from a given file, so add a general util function to handle this


caohe (approved):
lgtm

PR #47, opened 2023-04-25
feat(sysadvisor): support multi share regions

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
@@ -55,15 +55,19 @@ func init() {
 // 2. Support shared cores containers with different cpu enhancement
 // 3. Isolate bursting containers to isolation regions
 
-const (
-	startUpPeriod time.Duration = 30 * time.Second
-
-	regionNameSeparator = "-"
-)
-
 // InternalCalculationResult conveys minimal calculation result to cpu server
 type InternalCalculationResult struct {
 	PoolEntries map[string]map[int]resource.Quantity // map[poolName][numaId]cores
+}
+
+func (r *InternalCalculationResult) SetPoolEntry(poolName string, numaID int, poolSize int64) {
+	if poolSize <= 0 {
+		return
+	}
+	if r.PoolEntries[poolName] == nil {
+		r.PoolEntries[poolName] = make(map[int]resource.Quantity)
+	}
+	r.PoolEntries[poolName][numaID] = *resource.NewQuantity(poolSize, resource.DecimalSI)
 }
 
 // cpuResourceAdvisor is the entrance of updating cpu resource provision advice for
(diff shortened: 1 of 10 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -144,26 +145,32 @@ func (cra *cpuResourceAdvisor) GetHeadroom() (resource.Quantity, error) {
...
 		totalHeadroom.Add(headroom)
 	}
 
 	// Add headroom of numas without numa binding pods if there is no share region
-	if !hasShareRegion {
-		reservePoolSizeOfNonBindingNumas := int64(math.Ceil(float64(reservePoolSize*cra.nonBindingNumas.Size()) / float64(cra.metaServer.NumNUMANodes)))
-		headroomOfNonBindingNumas := resource.NewQuantity(int64(cra.nonBindingNumas.Size()*cra.metaServer.CPUsPerNuma())-reservePoolSizeOfNonBindingNumas, resource.DecimalSI)
-		totalHeadroom.Add(*headroomOfNonBindingNumas)
-	}
+	reservePoolSizeOfNonBindingNumas := int(math.Ceil(float64(reservePoolSize*cra.nonBindingNumas.Size()) / float64(cra.metaServer.NumNUMANodes)))
```
waynepeking348, line 169: reserve-size won't be adjusted during running process, so why don't we pre-calculate it, and store somewhere?

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -428,14 +437,48 @@ func (cra *cpuResourceAdvisor) assembleProvision() (InternalCalculationResult, e
 
 	// Fill in reclaimed pool size of non binding numas
 	reservePoolSizeOfNonBindingNumas := int(math.Ceil(float64(reservePoolSize*cra.nonBindingNumas.Size()) / float64(cra.metaServer.NumNUMANodes)))
-	reclaimPoolSizeOfNonBindingNumas := cra.nonBindingNumas.Size()*cra.metaServer.CPUsPerNuma() - nonNumaBindingRequirement - reservePoolSizeOfNonBindingNumas
+
+	reclaimPoolSizeOfNonBindingNumas := general.Max(cra.nonBindingNumas.Size()*cra.metaServer.CPUsPerNuma()-nonNumaBindingRequirement-reservePoolSizeOfNonBindingNumas,
```
waynepeking348, line 441: can we separate this function to several different functions? for instance, one to assemble dedicated pools, one for shared pools, one for reclaimed, and so on; just a proposal, you may raise up with a better separating way

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -428,14 +437,48 @@ func (cra *cpuResourceAdvisor) assembleProvision() (InternalCalculationResult, e
...
+	}
+
 	if provision.PoolEntries[state.PoolNameReclaim] == nil {
 		provision.PoolEntries[state.PoolNameReclaim] = make(map[int]resource.Quantity)
 	}
 	provision.PoolEntries[state.PoolNameReclaim][cpuadvisor.FakedNumaID] = *resource.NewQuantity(int64(reclaimPoolSizeOfNonBindingNumas), resource.DecimalSI)
 	return provision, nil
 }
 
+func genShareRegionPools(shareRegionRequirement map[string]int, sharePoolSize int) map[string]int {
```
waynepeking348, line 458: maybe this function should be moved below to make sure the method for cpuResourceAdvisor place together, just for a better coding style

waynepeking348: and it would be better if more comments are added for this function

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
cheney-lin: Although the size of the  pool is fixed, nonBindingNumas is dynamically changing, so we should maintain the logic here.

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -144,26 +145,32 @@ func (cra *cpuResourceAdvisor) GetHeadroom() (resource.Quantity, error) {
 		return *resource.NewQuantity(int64(cra.metaServer.NumCPUs-reservePoolSize), resource.DecimalSI), nil
 	}
 
-	hasShareRegion := false
+	shareRegionRequirement := 0
+
 	totalHeadroom := resource.NewQuantity(0, resource.DecimalSI)
 	for _, r := range cra.regionMap {
+		if r.Type() == types.QoSRegionTypeShare {
```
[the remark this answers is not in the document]
cheney-lin: I have had an in-depth discussion with @sun-yuliang  on this topic. If we want to decouple the calculation headroom of each share region, we need to set a specific total value for each region. However, how to divide the total value is a difficult problem. It may be possible to solve it through multiple iterations, but it is not a problem that we need to solve in the short term. That's why we wrote this rather ugly but logically correct code

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -520,3 +533,31 @@ func (cra *cpuResourceAdvisor) setPoolRegions(poolName string, regions []region.
 	}
 	return cra.metaCache.SetPoolInfo(poolName, pool)
 }
+
+// shareRegionRequirement refers to the resource demand of each share region,
+// sharePoolSize represents the total capacity of the share pool.
+// If the total resource demand of all share regions exceeds the total capacity of the pool,
+// we need to distribute the total capacity proportionally according to the resource requirement.
+func genShareRegionPools(shareRegionRequirement map[string]int, sharePoolSize int) map[string]int {
```
sun-yuliang, line 541: It seems that share pool sizes are changed in any case, while we only want to do that when the sum of share pool sizes exceeds their maximal allocatable resource. Also hard to understand why need to traverse map. Why not do something like this:

1. If sum of share pool sizes do not exceed non binding numa resources, do nothing.
2. Else restrict share pools in proportion as line 550 does, then return.

cheney-lin: 1. Already refined it this way.
2. If CPU is assigned to the multi share regions in proportion, the sum of share pool sizes may not be exactly equal to their maximal allocatable resource, so we need to make some adjustments.

sun-yuliang: If we have 569-571, it's more reasonable to remove line 568?

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -520,3 +540,37 @@ func (cra *cpuResourceAdvisor) setPoolRegions(poolName string, regions []region.
...
+	requirementSum := 0
+	for _, requirement := range shareRegionRequirement {
+		requirementSum += requirement
+	}
+	if requirementSum <= sharePoolSize {
+		return shareRegionRequirement
+	}
+
+	for poolName, requirement := range shareRegionRequirement {
+		sharePools[poolName] = sharePoolSize * requirement / requirementSum
```
sun-yuliang, line 560: Better to use float64 for requirement and requirementSum to avoid loss of precision

cheney-lin: Poolsize is an integer, so it is reasonable to use int

sun-yuliang: But requirement should be float64

sun-yuliang (dismissed):

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -55,17 +55,21 @@ func init() {
...
-
-	regionNameSeparator = "-"
-)
-
 // InternalCalculationResult conveys minimal calculation result to cpu server
 type InternalCalculationResult struct {
 	PoolEntries map[string]map[int]resource.Quantity // map[poolName][numaId]cores
 }
 
+func (r *InternalCalculationResult) SetPoolEntry(poolName string, numaID int, poolSize int64) {
```
waynepeking348, line 63: should this function can be exported?

sun-yuliang (approved):

waynepeking348 (approved):

PR #52, opened 2023-05-06
fix(sysadvisor): fix race between AddPod, RemovePod and RangeAndUpdat…

cheney-lin:
…eContainer

#### What type of PR is this?

Bug fixes


pkg/agent/sysadvisor/metacache/metacache.go
```diff
@@ -287,8 +288,18 @@ func (mc *MetaCacheImp) RangeAndUpdateContainer(f func(podUID string, containerN
 	}
 
 	mc.mutex.Lock()
-	mc.podEntries = podEntries
-	mc.mutex.Unlock()
+	defer mc.mutex.Unlock()
```
waynepeking348, line 291: add comments here: during range process, pod entries in metaCache may have already been deleted, so we will only update entries that still exists to avoid race condition

waynepeking348 (approved):

sun-yuliang (approved):

PR #57, opened 2023-05-09
feat(eviction manager): support pod-level rss overuse evict

zzzzhhb:
#### What type of PR is this?
Enhancements

support new policy to evict pods when pod use rss more than user expect.


go.mod
```diff
@@ -140,7 +140,7 @@ require (
 )
 
 replace (
-	github.com/kubewharf/katalyst-api => github.com/kubewharf/katalyst-api v0.1.2
+	github.com/kubewharf/katalyst-api => github.com/zzzzhhb/katalyst-api v0.0.0-20230519074205-64243545075a
```
waynepeking348, line 143: remember to switch to tags for katalyst api

pkg/util/qos/mem_enhancement.go
```diff
@@ -38,3 +40,28 @@ func IsPodNumaBinding(qosConf *generic.QoSConfiguration, pod *v1.Pod) bool {
...
+	// don't perform evict for safety if user set an unresolvable threshold
+	if parseErr != nil {
+		return false, consts.PodAnnotationMemoryEnhancementRssOverUseThresholdNotSet
+	}
+	// don't perform evict for safety if user set an invalid threshold
+	if !isValidRatioThreshold(threshold) {
+		return false, consts.PodAnnotationMemoryEnhancementRssOverUseThresholdNotSet
+	}
+
+	return true, threshold
```
waynepeking348, line 62: generally, we will return the boolean variables as the second one, maybe adjust will be better

pkg/agent/evictionmanager/plugin/memory/helper.go
```diff
@@ -0,0 +1,290 @@
...
+		})
+	}
+
+	return cmpFuncs
+
+}
+
+// getPodMetric returns the value of a pod-level metric.
+// And the value of a pod-level metric is calculated by summing the metric values for all containers in that pod.
+func (e *EvictionHelper) getPodMetric(pod *v1.Pod, metricName string, numaID int) (float64, bool) {
```
waynepeking348, line 242: we have a general function in metaserver, `AggregatePodNumaMetric`, can you use this to replace this self-defined logic here?

pkg/agent/evictionmanager/plugin/memory/helper.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	_ = e.emitter.StoreFloat64(metricsNameSystemMetric, kswapdSteal, metrics.MetricTypeNameRaw,
+		metrics.ConvertMapToTags(map[string]string{
+			metricsTagKeyMetricName: consts.MetricMemKswapdstealSystem,
+		})...)
+
+	return kswapdSteal, nil
+}
+
+func (e *EvictionHelper) selectPodsToEvict(activePods []*v1.Pod, topN uint64, numaID,
```
waynepeking348, line 166: since this logic if only used by system and numa plugin, and rss doesn't follow this. should we rename the function to clarify?

pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (r *RssOveruseEvictionPlugin) GetEvictPods(c context.Context, request *pluginapi.GetEvictPodsRequest) (*pluginapi.GetEvictPodsResponse, error) {
+	result := make([]*pluginapi.EvictPod, 0)
+
+	if !r.memoryEvictionPluginConfig.DynamicConf.EnableRssOveruseDetection() {
+		return &pluginapi.GetEvictPodsResponse{EvictPods: result}, nil
+	}
+
+	for i := range request.ActivePods {
```
waynepeking348, line 104: btw, let's limit this logic to shared_cores and reclaimed_cores only to avoid mis-eviction

pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
```diff
@@ -0,0 +1,149 @@
...
+		pod := request.ActivePods[i]
+		enabled, threshold := qos.GetRSSOverUseEvictEnabledAndThreshold(r.qosConf, pod)
+		if !enabled {
+			continue
+		}
+
+		if threshold == apiconsts.PodAnnotationMemoryEnhancementRssOverUseThresholdNotSet {
```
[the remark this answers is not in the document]
zzzzhhb: actually it will, because user might not set threshold explicitly.In this case, `threshold` will be `PodAnnotationMemoryEnhancementRssOverUseThresholdNotSet` too. I'll refactor it since this part of logic is hard to read.

waynepeking348 (dismissed):

pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
```diff
@@ -0,0 +1,149 @@
...
+			threshold = r.memoryEvictionPluginConfig.DynamicConf.RssOveruseRateThreshold()
+		}
+
+		var limit int64 = 0
+		var limitNotSet = false
+		for _, container := range pod.Spec.Containers {
+			containerMemLimit := container.Resources.Limits.Memory()
```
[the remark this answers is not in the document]
zzzzhhb: it seems like I can't use it directly because it can't tell if there is a container without memory limit.

pkg/agent/evictionmanager/plugin/memory/helper.go
zzzzhhb: the plugins  make different cmp functions according whether or not there is lost metrics. the `AggregatePodNumaMetric` won't return it for now. If we want reuse it here, we need to modify `AggregatePodNumaMetric`.

pkg/agent/evictionmanager/plugin/memory/helper.go
zzzzhhb: yeah, current name might be misleading.I'll rename it but I think the name should more tied with its functionality rather than with current callers. Maybe other caller will call it someday.

csfldf (approved):

PR #59, opened 2023-05-10
support utilization based canonical cpu headroom policy

luomingmeng:
#### What type of PR is this?
Features
To increase the utilization of reclaimed_cores pool, the cpu headroom report to cnr can be oversold to schedule more reclaimed_cores workload to the machine

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_adaptive.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_adaptive.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2022 The Katalyst Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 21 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_adaptive.go
```diff
@@ -0,0 +1,157 @@
...
+	p := &PolicyAdaptive{
+		PolicyBase:           NewPolicyBase(regionName, metaReader, metaServer, emitter),
+		policyAdaptiveConfig: conf.CPUHeadroomPolicyConfiguration.PolicyAdaptive,
+	}
+
+	return p
+}
+
+func (p *PolicyAdaptive) Update() error {
+	regionInfo, ok := p.metaReader.GetRegionInfo(p.regionName)
```
waynepeking348, line 54: why we need to get region-info and judge with its type? should we just verify this logic when NewPolicyAdaptive is called? for dedicated_cores, can it still work since it will never update() with nil error at all?

sun-yuliang (approved):

PR #62, opened 2023-05-11
fix(sysadvisor): clean up the contianers not existed in checkpoint

cheney-lin:
the `AddContainer` request may time out due to `storeState`, causing container leaks in metaCache, so it is necessary clean up any excess containers.

#### What type of PR is this?

Bug fixes


pkg/agent/sysadvisor/metacache/metacache.go
```diff
@@ -403,6 +406,14 @@ func (mc *MetaCacheImp) storeState() error {
 	checkpoint.PodEntries = mc.podEntries
 	checkpoint.PoolEntries = mc.poolEntries
 
+	begin := time.Now()
```
sun-yuliang, line 409: Do we need to log error when storing state exceeds time limit in any case? If so, move this hook function to store/restore state in metacache.

cheney-lin: Yes, this hook is already in storeState, and also called by restore

sun-yuliang: In storeState, exceeding time limit is not checked.

sun-yuliang (approved):

waynepeking348 (approved):

PR #63, opened 2023-05-11
fix(sysadvisor): fix failed to find regions and pools

cheney-lin:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
@@ -491,15 +491,42 @@ func (cra *cpuResourceAdvisor) gc() {
 	}
 }
 
-func (cra *cpuResourceAdvisor) getContainerRegions(ci *types.ContainerInfo) ([]region.QoSRegion, error) {
+func (cra *cpuResourceAdvisor) getRegionsByRegionNames(names sets.String) []region.QoSRegion {
 	var regions []region.QoSRegion = nil
-	for regionName := range ci.RegionNames {
+	for regionName := range names {
 		r, ok := cra.regionMap[regionName]
 		if !ok {
-			return nil, fmt.Errorf("failed to find region %v", regionName)
+			return nil
 		}
 		regions = append(regions, r)
 	}
+	return regions
+}
+
+func (cra *cpuResourceAdvisor) getRegionsByPodUID(podUID string) []region.QoSRegion {
+	var regions []region.QoSRegion = nil
+	for _, r := range cra.regionMap {
+		podSet := r.GetPods()
+		for uid := range podSet {
+			if uid == podUID {
+				regions = append(regions, r)
+			}
+		}
+	}
+	return regions
+}
+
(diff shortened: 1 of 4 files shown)


sun-yuliang (dismissed):

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/regulator/regulator.go
```diff
@@ -96,6 +96,7 @@ func (c *CPURegulator) slowdown(cpuRequirement float64) float64 {
 	// Restrict ramp down period
 	if cpuRequirement < latestCPURequirement && now.Before(c.latestRampDownTime.Add(c.minRampDownPeriod)) {
 		cpuRequirement = latestCPURequirement
```
waynepeking348, line 98: return latestCPURequirement; it seems we don't need to update to cpuRequirement

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -491,15 +491,39 @@ func (cra *cpuResourceAdvisor) gc() {
 	}
 }
 
-func (cra *cpuResourceAdvisor) getContainerRegions(ci *types.ContainerInfo) ([]region.QoSRegion, error) {
+func (cra *cpuResourceAdvisor) getRegionsByRegionNames(names sets.String) []region.QoSRegion {
 	var regions []region.QoSRegion = nil
-	for regionName := range ci.RegionNames {
-		r, ok := cra.regionMap[regionName]
+	for regionName := range names {
+		region, ok := cra.regionMap[regionName]
```
waynepeking348, line 497: region is the same with the import name, change this to r

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -491,15 +491,39 @@ func (cra *cpuResourceAdvisor) gc() {
...
+				regions = append(regions, r)
+			}
+		}
+	}
+	return regions
+}
+
+func (cra *cpuResourceAdvisor) getContainerRegions(ci *types.ContainerInfo) ([]region.QoSRegion, error) {
+	regions := cra.getRegionsByRegionNames(ci.RegionNames)
+	if len(regions) > 0 {
```
waynepeking348, line 521: when will `getRegionsByRegionNames` returns nil, but `getRegionsByPodUID` has data?

cheney-lin: When a new container(sidecar as example) allocated, its regionNames are empty, so `getRegionsByRegionNames` will returns nil. But its main continer has already assigned to specified region, so `getRegionsByPodUID` returns the regions of main container

waynepeking348: okay, got it, maybe we'd better add those above as comments to clarify

cheney-lin: done

waynepeking348 (approved):

sun-yuliang (approved):

PR #69, opened 2023-05-18
Dev/nic affinity

csfldf:
#### What type of PR is this?
feat(qrm): katalyst network qrm plugin supports nic affinitive allocation


pkg/agent/qrm-plugins/network/staticpolicy/types.go
```diff
@@ -0,0 +1,33 @@
...
+	IPv6 []net.IP
+}
+
+type NetworkInterface struct {
+	Name               string
+	AffinitiveNUMANode int
+	Enabled            bool
+	Addr               NetworkInterfaceAddr
+	NSAbsolutePath     string
+	NSName             string
```
smart2003, line 32: Shall we add the capacity and reservation to NIF as well? It is OK to add them later.

pkg/agent/qrm-plugins/network/staticpolicy/util.go
```diff
@@ -0,0 +1,500 @@
...
+		for _, dir := range dirs {
+			if !dir.IsDir() {
+				nsList = append(nsList, dir.Name())
+			}
+		}
+	}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	for _, ns := range nsList {
```
smart2003, line 254: I think we'd better filter out the "redundant" NICs here. Only those active host NICs are needed. For example, we can only allocate bandwidth on active physical NICs in a physical node.

pkg/util/machine/network_linux.go
```diff
@@ -0,0 +1,272 @@
...
+			// todo: to discuss if block agent start when any ns is parsed failed
+			if !conf.NetMultipleNS {
+				return nil, fmt.Errorf("failed to get network topology: %v", err)
+			}
+
+			general.Errorf("get network topology for ns %s failed: %v", ns, err)
+			continue
+		}
+
+		networkInfo.Interface = append(networkInfo.Interface, nicsInNs...)
```
csfldf, line 94: 1. Interface -> Interfaces may be better?  since there may be multi-nics in the machine 
2. why not use []*machine.InterfaceInfo type

waynepeking348 (approved):

luomingmeng (approved):

PR #72, opened 2023-05-21
export external managers to SetNetworkManager

waynepeking348:
#### What type of PR is this?
bug fixes

export external managers to SetNetworkManager


caohe (approved):
LGTM

PR #89, opened 2023-06-04
Support isolation pools

waynepeking348:
#### What type of PR is this?
features

support isolation based on container load in sysadvisor


sun-yuliang (approved):

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor_helper.go
```diff
@@ -191,8 +192,18 @@ func (cra *cpuResourceAdvisor) getRegionMaxRequirement(regionName string) float6
 	}
 
 	res := 0.0
-	for _, numaID := range r.GetBindingNumas().ToSliceInt() {
-		res += float64(cra.numaAvailable[numaID])
+	switch r.Type() {
+	case types.QoSRegionTypeIsolation:
```
cheney-lin, line 196: I don't quite understand why only limits of main containers are considered here.

PR #98, opened 2023-06-07
fix: avoid fetching empty metrics

cheney-lin:
As metricsFetcher.GetContainerMetric and cpuResourceAdvisor.update are called in separate goroutines, the latter may be executed before the former, resulting in obtaining empty metrics and falling back to using pod requests as estimated resources, which leads to a sharp decrease in headroom.

#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
@@ -155,6 +155,12 @@ func (cra *cpuResourceAdvisor) update() {
 	cra.mutex.Lock()
 	defer cra.mutex.Unlock()
 
+	// skip updating during startup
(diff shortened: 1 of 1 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/advisor.go
```diff
@@ -152,6 +152,11 @@ func (cra *cpuResourceAdvisor) GetHeadroom() (resource.Quantity, error) {
 // update works in a monolithic way to maintain lifecycle and triggers update actions for all regions;
 // todo: re-consider whether it's efficient or we should make start individual goroutine for each region
 func (cra *cpuResourceAdvisor) update() {
+	// skip updating during startup
```
waynepeking348, line 155: should we put this under lock() ?

sun-yuliang (approved):

waynepeking348 (approved):

Issue #85, opened 2023-05-31
refine resource advisors in sysadvisor


sun-yuliang:
### What would you like to be added?

refine resource advisors, especially cpu advisor in sysadvisor

### Why is this needed?

1. several bugs exist in cpu advisor, may result in faulty provision and headroom result in certain cases
2. update logic in cpu advisor is complicated, more abstraction and simplification is required

PR #99, opened 2023-06-08
exchange round and slowdown in cpu regulator

sun-yuliang:
#### What type of PR is this?
Enhancements
-->

exchange round and slowdown procedure in cpu regulator to ignore trivial ramp down of float number

#85


waynepeking348 (approved):

csfldf (approved):

sun-yuliang:
> What are the benefits of this pr?

Under former logic, it's common that cpu requirement float number decreases a bit, i.e. same after ceil to even number, but triggers ramp down period update, resulting in ramp down too slow in some cases. After this exchange, this issue is solved.

PR #105, opened 2023-06-13
implement rama policy framework with pid controller

sun-yuliang:
#### What type of PR is this?
Features

Implement the framework of a fine tuned policy "rama" for dedicated numa exclusive region


pkg/agent/sysadvisor/types/types.go
```diff
@@ -231,3 +238,13 @@ const (
 	PolicyUpdateSucceeded PolicyUpdateStatus = "succeeded"
 	PolicyUpdateFailed    PolicyUpdateStatus = "failed"
 )
+
+// FirstOrderPIDParams holds parameters for pid controller in rama policy
+type FirstOrderPIDParams struct {
```
waynepeking348, line 243: should we use camel case instead of snake case?

pkg/agent/sysadvisor/plugin/qosaware/resource/helper/pid.go
```diff
@@ -0,0 +1,89 @@
...
+		controlKnobPrev: 0,
+		errorValue:      0,
+		errorValuePrev:  0,
+		errorRate:       0,
+	}
+}
+
+func (c *PIDController) Adjust(controlKnob, target, current float64) float64 {
+	var (
+		kp, kd, kp_sign, kd_sign float64
```
waynepeking348, line 48: Use camel case instead of snake case

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
```diff
@@ -86,3 +81,54 @@ func (r *QoSRegionDedicatedNumaExclusive) TryUpdateProvision() {
...
+
+	// get cpi of the main container in this region
+	for podUID, containerSet := range r.podSet {
+		for containerName := range containerSet {
+			ci, ok := r.metaReader.GetContainerInfo(podUID, containerName)
+			if !ok || ci == nil {
+				klog.Errorf("[qosaware-cpu] illegal container info of %v/%v", podUID, containerName)
+				return nil
+			}
+			if ci.ContainerType == v1alpha1.ContainerType_MAIN {
```
waynepeking348, line 116: this coding style is kind of wired: it assumes that that there should be only one valid pod/container (the latter cpi will always override the previous one); but there is nowhere to limit this, so, i'd rather to sum up the cpi for all those pod/containers in this region

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
```diff
@@ -86,3 +81,54 @@ func (r *QoSRegionDedicatedNumaExclusive) TryUpdateProvision() {
...
+					klog.Errorf("[qosaware-cpu] get %v of %v/%v failed: %v", consts.MetricCPUCPIContainer, podUID, containerName, err)
+					return nil
+				}
+			}
+		}
+	}
+
+	return types.Indicator{
+		consts.MetricCPUCPIContainer: {
+			Current: cpi,
```
waynepeking348, line 128: and, should we emit error if there is no valid pod or container (i.e, cpi = 0 or something like that)?

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/provisionpolicy/policy_rama.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// 3. check pid controllers
+	for _, metricName := range p.configuredIndicatorMetrics {
+		if _, ok := p.controllers[metricName]; !ok {
+			errList = append(errList, fmt.Errorf("missing controller for indicator %v", metricName))
+		}
+	}
+
+	// 4. check margin. skip update when margin is non zero
+	if p.ResourceEssentials.ReservedForAllocate != 0 {
```
waynepeking348, line 141: if we can't set reservation for rama, how can we do to relief the influence for online service?

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/provisionpolicy/policy_rama.go
```diff
@@ -0,0 +1,146 @@
...
+		} else if indicator.Target <= 0 || indicator.Current <= 0 {
+			isLegal = false
+		}
+	}
+	if !isLegal {
+		errList = append(errList, fmt.Errorf("illegal indicator %v", p.Indicators))
+	}
+
+	// 3. check pid controllers
+	for _, metricName := range p.configuredIndicatorMetrics {
```
waynepeking348, line 134: will it be better to return error during new functions? since this configuredIndicatorMetrics will always return error here, and will never works as expected anymore

waynepeking348:
there still exists some comments, we need to discuss before this pr can be merged

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_canonical.go
```diff
@@ -17,16 +17,14 @@ limitations under the License.
 package headroompolicy
```
waynepeking348, line 17: if this policy is deprecated, which one should we use? will there be a corresponding rama headroom policy?

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
sun-yuliang: Policy will handle all kinds of abnormal cases and write error logs

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/provisionpolicy/policy_rama.go
sun-yuliang: Once reservation is set, rama policy will fail and advisor will use canonical policy result as fallback.

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/provisionpolicy/policy_rama.go
sun-yuliang: PID controller pamameters may be dynamically changed through KCC in the future so we need dynamic check here.

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_canonical.go
sun-yuliang: There may be a policy based on machine learning to estimate NUMA headroom for dedicated numa exclusive region in the future.

cheney-lin (approved):

PR #106, opened 2023-06-15
feat(advisor_proto): design a general api between qrm plugins and advisors

csfldf:
#### What type of PR is this?

feat(advisor_proto): design a general api between qrm plugins and advisors


waynepeking348 (approved):

pkg/agent/qrm-plugins/advisorsvc/advisor_svc.proto
```diff
@@ -0,0 +1,85 @@
...
+}
+
+message RemovePodRequest {
+        string pod_uid = 1;
+}
+
+message RemovePodResponse {
+}
+
+message ListAndWatchResponse {
```
cheney-lin, line 61: Is this message only for memory plugin/advisor ?

csfldf: It's not designed for memory. If there are new plugins and advisors can reuse this api, they can reuse it directly.

cheney-lin (approved):

PR #112, opened 2023-06-19
calculate container cpi by cycles and instructions

sun-yuliang:
#### What type of PR is this?
Bug fix

Calculate container cpi metric by delta(cycles)/delta(instructions).

### DIFF
--- a/pkg/metaserver/agent/metric/metric.go
+++ b/pkg/metaserver/agent/metric/metric.go
@@ -457,8 +457,12 @@ func (m *MalachiteMetricsFetcher) processSystemCPUComputeData(systemComputeData 
 func (m *MalachiteMetricsFetcher) processCgroupCPUData(podUID, containerName string, cgStats *cgroup.MalachiteCgroupInfo) {
 	m.processContainerMemBandwidth(podUID, containerName, cgStats)
 
+	cyclesOld, _ := m.GetContainerMetric(podUID, containerName, consts.MetricCPUCyclesContainer)
+	instructionsOld, _ := m.GetContainerMetric(podUID, containerName, consts.MetricCPUInstructionsContainer)
+
 	if cgStats.CgroupType == "V1" {
 		cpu := cgStats.V1.Cpu
+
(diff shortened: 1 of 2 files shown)


pkg/metaserver/agent/metric/metric.go
```diff
@@ -134,6 +134,9 @@ func NewMalachiteMetricsFetcher(emitter metrics.MetricEmitter) MetricsFetcher {
 }
 
 type MalachiteMetricsFetcher struct {
+	cpuCgDataV1 cgroup.CPUCgDataV1
```
waynepeking348, line 137: if we store this here, why we still need metricStore? it's wired to just keep cpuCgDataV1/cpuCgDataV2 here, why not get previous Instructions from  metricStore?

sun-yuliang: fixed

waynepeking348 (approved):

pkg/metaserver/agent/metric/metric.go
```diff
@@ -584,9 +614,9 @@ func (m *MalachiteMetricsFetcher) processCgroupPerfData(podUID, containerName st
 	if perf == nil {
 		return
 	}
-	m.metricStore.SetContainerMetric(podUID, containerName, consts.MetricCPUCPIContainer, perf.Cpi)
-	m.metricStore.SetContainerMetric(podUID, containerName, consts.MetricCPUCyclesContainer, perf.Cycles)
-	m.metricStore.SetContainerMetric(podUID, containerName, consts.MetricCPUInstructionsContainer, perf.Instructions)
+	// m.metricStore.SetContainerMetric(podUID, containerName, consts.MetricCPUCPIContainer, perf.Cpi)
```
cheney-lin, line 617: It's not a good idea to use comments to delete code.

sun-yuliang: fixed

waynepeking348 (approved):

PR #124, opened 2023-06-28
add headroom policy and assembler for dedicated

sun-yuliang:
#### What type of PR is this?
Enhancements

Add a headroom policy based on load and headroom assembler for scene with dedicated containers.

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/assembler.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/assembler.go
@@ -27,13 +27,18 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/config"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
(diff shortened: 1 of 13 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/assembler.go
```diff
@@ -27,13 +27,15 @@ import (
...
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
+	"github.com/kubewharf/katalyst-core/pkg/util/machine"
 )
 
 type HeadroomAssembler interface {
 	GetHeadroom() (resource.Quantity, error)
 }
 
 type InitFunc func(conf *config.Configuration, extraConf interface{}, regionMap *map[string]region.QoSRegion,
+	reservedForReclaim *map[int]int, numaAvailable *map[int]int, nonBindingNumas *machine.CPUSet,
```
waynepeking348, line 38: why should we use map pointer here? does it mean that we will depend on this pointer to get the latest version for the map (which is not proper actually)? why not put those info in metacache?

cheney-lin (approved):

cheney-lin (approved):

PR #125, opened 2023-06-28
enhancement(eviction): avoid misjudgment by checking metric timestamp

zzzzhhb:
#### What type of PR is this?

Enhancements

system pressure eviction plugin use the diff value of kswapdsteal metric between two rounds of sync to check if the system is under memory pressure. But sometimes the metric is not updated as in time as we expected, it will lead to misjudgment. This PR checks the metric timestamp to prevent that misjudgment. When the metric timestamp is same with it in last round, we ignore this round.

### DIFF
--- a/pkg/agent/evictionmanager/plugin/memory/helper.go
+++ b/pkg/agent/evictionmanager/plugin/memory/helper.go
@@ -161,10 +161,14 @@ func (e *EvictionHelper) getWatermarkMetrics(numaID int) (free, total, scaleFact
 	return free, total, scaleFactor, nil
 }
 
-func (e *EvictionHelper) getSystemKswapdStealMetrics() (float64, error) {
+func (e *EvictionHelper) getSystemKswapdStealMetrics() (metric.MetricData, error) {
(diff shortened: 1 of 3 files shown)


pkg/agent/evictionmanager/plugin/memory/helper.go
```diff
@@ -161,10 +161,10 @@ func (e *EvictionHelper) getWatermarkMetrics(numaID int) (free, total, scaleFact
 	return free, total, scaleFactor, nil
 }
 
-func (e *EvictionHelper) getSystemKswapdStealMetrics() (float64, error) {
+func (e *EvictionHelper) getSystemKswapdStealMetrics() (*metric.MetricData, error) {
 	m, err := e.metaServer.GetNodeMetric(consts.MetricMemKswapdstealSystem)
 	if err != nil {
```
waynepeking348, line 166: if you want to use m.Time, maybe it's better to check whether it's nil here

zzzzhhb: updated

waynepeking348 (approved):

luomingmeng (approved):

PR #135, opened 2023-07-08
feat(metaserver): support collecting metrics for qos level

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/metaserver/agent/metric/metric.go
+++ b/pkg/metaserver/agent/metric/metric.go
@@ -29,11 +29,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
+	"github.com/kubewharf/katalyst-core/pkg/config"
 	"github.com/kubewharf/katalyst-core/pkg/consts"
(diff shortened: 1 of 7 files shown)


pkg/metaserver/agent/metric/metric.go
```diff
@@ -287,6 +306,23 @@ func (m *MalachiteMetricsFetcher) updateSystemStats() {
 	}
 }
 
+func (m *MalachiteMetricsFetcher) updateQoSClassCgroupData() {
+	qosClassCgroupPath := []string{m.conf.ReclaimRelativeRootCgroupPath, common.CgroupFsRootPathBurstable, common.CgroupFsRootPathBestEffort}
```
waynepeking348, line 310: the name `qosClass` in this file is kind of confusing, it represents both `k8s qos` and `katalyst qos`,   will it be more intuitionistic if we just name the function as `GetCgroupMetric` and `GetQoSClassNumaMetric`, and the parameter as `cgroup` instead of `qosClass`?

cheney-lin: done

waynepeking348 (approved):

sun-yuliang (approved):

PR #143, opened 2023-07-12
memory advisor plugins

cheney-lin:
#### What type of PR is this?

Features

### DIFF
--- a/pkg/agent/evictionmanager/plugin/memory/numa_pressure.go
+++ b/pkg/agent/evictionmanager/plugin/memory/numa_pressure.go
@@ -31,6 +31,7 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/config"
 	"github.com/kubewharf/katalyst-core/pkg/config/agent/dynamic"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
+	"github.com/kubewharf/katalyst-core/pkg/metaserver/agent/metric/helper"
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
 	"github.com/kubewharf/katalyst-core/pkg/util/general"
 	"github.com/kubewharf/katalyst-core/pkg/util/native"
@@ -123,7 +124,7 @@ func (n *NumaMemoryPressurePlugin) detectNumaPressures() {
 }
 
 func (n *NumaMemoryPressurePlugin) detectNumaWatermarkPressure(numaID int) error {
-	free, total, scaleFactor, err := n.evictionHelper.getWatermarkMetrics(numaID)
+	free, total, scaleFactor, err := helper.GetWatermarkMetrics(n.metaServer.MetricsFetcher, n.emitter, numaID)
 	if err != nil {
 		general.Errorf("failed to getWatermarkMetrics for numa %d, err: %v", numaID, err)
 		_ = n.emitter.StoreInt64(metricsNameFetchMetricError, 1, metrics.MetricTypeNameCount,
--- a/pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
+++ b/pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/config/generic"
 	"github.com/kubewharf/katalyst-core/pkg/consts"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
+	"github.com/kubewharf/katalyst-core/pkg/metaserver/agent/metric/helper"
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
 	"github.com/kubewharf/katalyst-core/pkg/util/general"
 	"github.com/kubewharf/katalyst-core/pkg/util/process"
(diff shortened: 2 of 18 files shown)


pkg/agent/sysadvisor/metacache/metacache.go
```diff
@@ -46,6 +46,8 @@ const (
 
 // MetaReader provides a standard interface to refer to metadata type
 type MetaReader interface {
+	// GetContainers returns containers filtered by containerFilter
+	GetContainers(containerFilter func(*types.ContainerInfo) bool) []*types.ContainerInfo
```
waynepeking348, line 50: it is unnecessary to add this function since it can be totally achieved by `RangeContainer`

pkg/agent/evictionmanager/plugin/memory/numa_pressure.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/client"
```
waynepeking348, line 30: why we must modify so much for eviction logic? we have currently made it online running for this logic, so if you want to modify this, please contact with @zzzzhhb and make sure you have fully tested (compared with current version) and access permission from @zzzzhhb

pkg/util/qos/qos_helper.go
```diff
@@ -0,0 +1,183 @@
+/*
```
waynepeking348, line 1: it has nothing to do with qos at all; if you want to make this logic more general, you should move it to dir `util/metic` and arrange them into several files like `metric_numa.go`, `metric_pod.go` or something like that

pkg/consts/metric.go
```diff
@@ -231,3 +231,29 @@ const (
 	MetricsMemFilePerNumaCgroup  = "mem.file.numa.cgroup"
 	MetricsMemAnonPerNumaCgroup  = "mem.anon.numa.cgroup"
 )
+
+const (
```
waynepeking348, line 235: this defines metric rather than metrics (to emit); actually, we should the metrics in wherever requires them, and no need to define them in a general place

pkg/agent/sysadvisor/plugin/qosaware/resource/memory/advisor.go
```diff
@@ -33,17 +34,24 @@ import (
...
+	"github.com/kubewharf/katalyst-core/pkg/consts"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
 	"github.com/kubewharf/katalyst-core/pkg/util/general"
+	"github.com/kubewharf/katalyst-core/pkg/util/qos"
 )
 
 func init() {
 	headroompolicy.RegisterInitializer(types.MemoryHeadroomPolicyCanonical, headroompolicy.NewPolicyCanonical)
+	memadvisorplugin.RegisterInitializer(memadvisorplugin.CacheReaper, memadvisorplugin.NewCacheReaper)
```
waynepeking348, line 46: add a blank to separate different logic (policy and plugin)

pkg/agent/evictionmanager/plugin/memory/numa_pressure.go
cheney-lin: These modifications do not involve functional changes, but only extract common public functions.

pkg/util/qos/qos_helper.go
cheney-lin: If it is defined in util/metic, it will introduce circular reference problems, so it has to be placed in another package.

pkg/consts/metric.go
cheney-lin: They are defined here because they are referenced by eviction and memory advisor.

waynepeking348: they should not be defined here, the reason is as above

pkg/agent/evictionmanager/plugin/memory/numa_pressure.go
waynepeking348: you should confirm this with @zzzzhhb

pkg/util/qos/qos_helper.go
waynepeking348: maybe it can be defined in `metaserver/agent/metric/util`, and to avoid circular reference, the function should better use `MetricsFetcher` rather than `metaServer` as paramethers

pkg/consts/metric.go
waynepeking348: generally, we don't encourage different components to share the same metric names

zzzzhhb: maybe we should use different name for these two "metrics", it‘s really easy to get confused. For example:
runtime metric: the information about running containers, pods and system;
metric: the data we used to monitor katalyst itself.

for this certain case, I agree with @waynepeking348 . Maybe someday we will deploy advisor and eviction manager seperately and they might be maintained by different people. If they use same metrics name, it will be another confusing thing because it's not only reflect the status about one component.

waynepeking348 (approved):

pkg/agent/evictionmanager/plugin/memory/rss_overuse.go
```diff
@@ -164,7 +165,7 @@ func (r *RssOveruseEvictionPlugin) GetEvictPods(_ context.Context, request *plug
 			continue
 		}
 
-		podRss, found := r.evictionHelper.getPodMetric(pod, consts.MetricMemRssContainer, nonExistNumaID)
```
zzzzhhb, line 167: it seems like evictionHelper is useless in this plugin, maybe we can delete it.

waynepeking348 (approved):

zzzzhhb (approved):

PR #158, opened 2023-07-19
fix(sysadvisor): set numa capicity as requirement of numabinding pods  when reclaimed disabled

cheney-lin:
… when reclaimed disabled

#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_canonical.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_canonical.go
@@ -21,6 +21,7 @@ import (
 	"math"
 
 	"k8s.io/klog/v2"
+	"k8s.io/kubelet/pkg/apis/resourceplugin/v1alpha1"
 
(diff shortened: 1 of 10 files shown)


sun-yuliang (approved):

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/headroompolicy/policy_canonical.go
```diff
@@ -55,6 +55,14 @@ func (p *PolicyCanonical) Update() error {
 			return err
 		}
 
+		if p.regionType == types.QoSRegionTypeDedicatedNumaExclusive && !enableReclaim {
```
waynepeking348, line 58: should this be moved out of range loop for `for podUID, containerSet := range p.podSet {`? since you will add  up multiple times for `cpuEstimation` and `containerCnt` this way

waynepeking348: and the same for `provision_canonical.go`

csfldf (approved):

waynepeking348 (approved):

sun-yuliang (approved):

PR #171, opened 2023-07-26
fix(qrm): cpu plugin allocate for blocks according to calculated sequence

csfldf:
fix(qrm): cpu plugin allocate for blocks according to calculated sequence


pkg/agent/qrm-plugins/cpu/dynamicpolicy/cpuadvisor/helper.go
```diff
@@ -39,23 +42,18 @@ func NewBlockCPUSet() BlockCPUSet {
...
-	for _, cpuset := range b {
-		cpusets = cpusets.Union(cpuset)
-	}
-	return cpusets
-}
-
 // GetBlocks parses ListAndWatchResponse and returns map[int]map[string]*Block,
 // the map is keyed as numa id -> block id -> block
-func (lwr *ListAndWatchResponse) GetBlocks() (map[int]map[string]*Block, error) {
+func (lwr *ListAndWatchResponse) GetBlocks() (map[int]map[string]*Block, map[int][]string, error) {
```
waynepeking348, line 47: why should still keep the previous returned value (i.e. `map[int]map[string]*Block`) here? just for legacy reasons or for coding convenience ?

waynepeking348 (approved):

luomingmeng (approved):

luomingmeng (approved):

waynepeking348 (approved):

PR #178, opened 2023-08-02
fix unit test race in config manager

luomingmeng:
#### What type of PR is this?

Bug fixes
fix unit test race in config manager


pkg/metaserver/config/checkpoint.go
```diff
@@ -37,7 +37,7 @@ type ConfigManagerCheckpoint interface {
 type TargetConfigData struct {
 	// Value only store spec of dynamic config crd
 	Value     *crd.DynamicConfigCRD
-	Timestamp metav1.Time
+	Timestamp int64
```
waynepeking348, line 40: why modify this can help with race conditions?

caohe (approved):

PR #195, opened 2023-08-09
fix(qrm): return an error if no available nics on a node

smart2003:
#### What type of PR is this?
Bug fixes

 return an error if no available nics on a node

panic if no available nics on a node

### DIFF
--- a/pkg/agent/qrm-plugins/network/staticpolicy/policy.go
+++ b/pkg/agent/qrm-plugins/network/staticpolicy/policy.go
@@ -30,7 +30,6 @@ import (
 	maputil "k8s.io/kubernetes/pkg/util/maps"
 
 	apinode "github.com/kubewharf/katalyst-api/pkg/apis/node/v1alpha1"
-	"github.com/kubewharf/katalyst-api/pkg/consts"
 	apiconsts "github.com/kubewharf/katalyst-api/pkg/consts"
 
 	"github.com/kubewharf/katalyst-api/pkg/plugins/skeleton"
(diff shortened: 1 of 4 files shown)


pkg/agent/qrm-plugins/network/staticpolicy/policy.go
```diff
@@ -99,6 +99,9 @@ func NewStaticPolicy(agentCtx *agent.GenericContext, conf *config.Configuration,
 	// it is incorrect to reserve bandwidth on those diabled NICs.
 	// we only count active NICs as available network devices and allocate bandwidth on them
 	enabledNICs := filterNICsByAvailability(agentCtx.KatalystMachineInfo.ExtraNetworkInfo.Interface, nil, nil)
+	if len(enabledNICs) == 0 {
+		return false, agent.ComponentStub{}, fmt.Errorf("no available nics on this node")
```
csfldf, line 103: it may cause the plugin restarts repeatedly,  if we should make adaption when enabledNICs is empty instead of returning error?

csfldf (approved):

waynepeking348 (approved):

pkg/agent/qrm-plugins/network/staticpolicy/policy_test.go
```diff
@@ -1167,137 +1337,276 @@ func TestGetResourcesAllocation(t *testing.T) {
...
+			assert.Equal(t, resp.AllocatableResources[string(consts.ResourceNetBandwidth)].TopologyAwareCapacityQuantityList, tc.expectedCapacityQuantityList)
+		} else {
+			assert.Equal(t, float64(0), resp.AllocatableResources[string(consts.ResourceNetBandwidth)].AggregatedAllocatableQuantity)
+			assert.Equal(t, float64(0), resp.AllocatableResources[string(consts.ResourceNetBandwidth)].AggregatedCapacityQuantity)
+			assert.Len(t, resp.AllocatableResources[string(consts.ResourceNetBandwidth)].TopologyAwareAllocatableQuantityList, 0)
+			assert.Len(t, resp.AllocatableResources[string(consts.ResourceNetBandwidth)].TopologyAwareCapacityQuantityList, 0)
+		}
+	}
+
+	/*
```
waynepeking348, line 1542: we should delete all the unless codes instead of using comments

smart2003: Oh, it was left by mistake. Just forgot it.

smart2003: Just removed

waynepeking348 (approved):

pkg/agent/qrm-plugins/network/staticpolicy/policy.go
```diff
@@ -580,13 +584,42 @@ func (p *StaticPolicy) Allocate(_ context.Context,
...
 
 	if len(candidateNICs) == 0 {
-		general.ErrorS(err, "insufficient bandwidth on this node to satisfy the request",
-			"podNamespace", req.PodNamespace,
-			"podName", req.PodName,
-			"containerName", req.ContainerName,
-			"netBandwidthReq(Mbps)", reqInt,
-			"nicState", p.state.GetMachineState().String())
-		return nil, fmt.Errorf("failed to meet the bandwidth requirement of %d Mbps", reqInt)
+		if reqInt > 0 {
```
csfldf, line 587: Remains original codes (return err if len(candidateNICs) == 0) will be ok. 

Since for some pod reqInt == 0, but we should allocate affinitive netcard for it. If there is no available nics, we should return err in this condition.

csfldf (approved):

waynepeking348 (approved):

PR #200, opened 2023-08-10
feat(sysadvisor): add indicator cpu usage ratio in rama policy for share region

sun-yuliang:
#### What type of PR is this?
Enhancements

Add indicator cpu usage ratio in rama policy for share region to restrict share pool cpu usage.


pkg/metaserver/agent/metric/malachite/fetcher.go
```diff
@@ -474,8 +474,8 @@ func (m *MalachiteMetricsFetcher) processSystemCPUComputeData(systemComputeData
 			klog.Errorf("[malachite] parse cpu name %v with err: %v", cpu.Name, err)
 			continue
 		}
-		m.metricStore.SetCPUMetric(cpuID, consts.MetricCPUUsage,
-			utilmetric.MetricData{Value: cpu.CPUUsage, Time: &updateTime})
+		m.metricStore.SetCPUMetric(cpuID, consts.MetricCPUUsageRatio,
```
waynepeking348, line 477: maybe it's better to keep both those two metrics, since we will depend on the previous metric in other cases

waynepeking348 (approved):

csfldf (approved):

pkg/metaserver/agent/metric/malachite/fetcher.go
sun-yuliang: CPUUsage will be changed to CPUUsageRatio globally in all cases

PR #208, opened 2023-08-15
fix(sysadvisor): replace memory avaiable metric

cheney-lin:
If the value of "high_wmark_pages" in the system is large, "memory available" metric will be too low even zero from /proc/meminfo. We use "free+cache+buffer" as a substitute to figure this issue out.

#### What type of PR is this?

Bug fixes


pkg/metaserver/agent/metric/malachite/fetcher.go
```diff
@@ -714,6 +716,7 @@ func (m *MalachiteMetricsFetcher) processContainerCPUData(podUID, containerName
 				cpi := (float64(cpu.Cycles) - cyclesOld.Value) / instructionDiff
 				m.metricStore.SetContainerMetric(podUID, containerName, consts.MetricCPUCPIContainer,
 					utilmetric.MetricData{Value: cpi, Time: &updateTime})
+				m.synced = true
```
waynepeking348, line 719: why should we set synced here?

cheney-lin: typo

waynepeking348 (approved):

csfldf (approved):

PR #233, opened 2023-08-23
some rand functions are deprecated

zhy76:
1. rand.Seed:
Deprecated: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access.
2. rand.Read
Deprecated: For almost all use cases, crypto/rand.Read is more appropriate.

### DIFF
--- a/cmd/katalyst-agent/main.go
+++ b/cmd/katalyst-agent/main.go
@@ -18,9 +18,7 @@ package main
 
 import (
 	"fmt"
-	"math/rand"
 	"os"
-	"time"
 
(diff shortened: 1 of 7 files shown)


waynepeking348 (approved):

caohe:
@luomingmeng could u PTAL?

luomingmeng (approved):

PR #250, opened 2023-08-31
fix(controller): fix cnr residue when node deletion

chenxi-seu:
#### What type of PR is this?

Bug fixes

When a Node is deleted, there may be remaining CNR (CustomNodeResource) objects without labels in cluster. This is because when a node is deleted, katalyst-agent creates a new CNR without labels, and the katalyst-controller is unable to GET the deleted node for patching labels. 
Additionally, during the periodic cleaning process in katalyst-controller, it lists CNRs with labels based on the NodeSelector from Informer initialization, excluding the CNR without labels created katalyst-agent. This causes the CNR to be unable to be cleaned up.
```
# Residual CNR without lables
apiVersion: node.katalyst.kubewharf.io/v1alpha1
kind: CustomNodeResource
metadata:
  name: 172.25.0.7
spec: {}
status: 
  topologyPolicy: SingleNUMANodeContainerLevel
  topologyZone:
     ...
  
```

This PR attempts to list all CNRs from the kube-apiserver during the periodic cleaning process, instead of using `cnrLister` with a NodeSelector.

### DIFF
--- a/pkg/agent/resourcemanager/reporter/cnr/cnrreporter.go
+++ b/pkg/agent/resourcemanager/reporter/cnr/cnrreporter.go
@@ -61,6 +61,8 @@ const (
 type cnrReporterImpl struct {
 	cnrName string
 
+	// defaultLabels contains the default config for CNR created by reporter
+	defaultLabels map[string]string
 	// latestUpdatedCNR is used as an in-memory cache for CNR;
 	// whenever CNR info is needed, get from this cache firstly
 	latestUpdatedCNR *nodev1alpha1.CustomNodeResource
@@ -83,6 +85,7 @@ func NewCNRReporter(genericClient *client.GenericClientSet, metaServer *metaserv
 	c := &cnrReporterImpl{
 		cnrName:                conf.NodeName,
 		refreshLatestCNRPeriod: conf.RefreshLatestCNRPeriod,
+		defaultLabels:          conf.DefaultCNRLabels,
 		notifiers:              make(map[string]metaservercnr.CNRNotifier),
 		emitter:                emitter,
 		client:                 genericClient.InternalClient,
@@ -337,7 +340,8 @@ func (c *cnrReporterImpl) resetCNRIfNeeded(err error) bool {
 func (c *cnrReporterImpl) defaultCNR() *nodev1alpha1.CustomNodeResource {
 	return &nodev1alpha1.CustomNodeResource{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: c.cnrName,
+			Name:   c.cnrName,
+			Labels: c.defaultLabels,
 		},
 	}
 }
--- a/pkg/controller/lifecycle/cnr.go
+++ b/pkg/controller/lifecycle/cnr.go
@@ -320,7 +320,7 @@ func (cl *CNRLifecycle) updateOrCreateCNR(node *corev1.Node) error {
 	}
 
 	newCNR := cnr.DeepCopy()
-	newCNR.Labels = node.Labels
+	newCNR.Labels = general.MergeMap(newCNR.Labels, node.Labels)
 	setCNROwnerReference(newCNR, node)
 	if apiequality.Semantic.DeepEqual(newCNR, cnr) {
 		return nil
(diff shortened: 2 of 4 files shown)


waynepeking348:
maybe it's better to do this by add the same label in `reporter` rather than trigger `list` request periodically cc @caohe @luomingmeng

chenxi-seu:
@waynepeking348 Sorry for the late reply. The method you mentioned indeed has better performance. I will submit the PR later.

chenxi-seu:
@waynepeking348 @caohe PTAL
1. add `--label-selector` to the agent startup parameter, which controls the default labels of CNR created by agent. User needs to ensure that this parameter is consistent with the `--label-selector` for controller. For example, use the same value in Charts `values.yaml` to render the label-selector for both agent and controller.
2. merge labels instead of overriding it with node lables when controller periodic sync CNR.

pkg/controller/lifecycle/cnr.go
```diff
@@ -347,3 +347,9 @@ func setCNROwnerReference(cnr *apis.CustomNodeResource, node *corev1.Node) {
 		},
 	}
 }
+
+func mergeLabels(source, destination map[string]string) {
```
waynepeking348, line 351: there exists an function `MergeMap` in util/general.go, maybe it's better to use that instead of defining it again here

cmd/katalyst-agent/app/options/global/base.go
```diff
@@ -27,6 +27,7 @@ type BaseOptions struct {
 	Agents             []string
 	NodeName           string
 	NodeAddress        string
+	LabelSelector      string
```
luomingmeng, line 30: This option/config should be renamed to `DefaultCNRLabels` and moved GenericReporterOptions/GenericReporterConfiguration, which may be more properly.

pkg/controller/lifecycle/cnr.go
chenxi-seu: ok, I will replace it with the general function.

cmd/katalyst-agent/app/options/global/base.go
```diff
@@ -55,6 +56,8 @@ func (o *BaseOptions) AddFlags(fss *cliflag.NamedFlagSets) {
 	fs.StringSliceVar(&o.Agents, "agents", o.Agents, "The agents need to be started")
 	fs.StringVar(&o.NodeName, "node-name", o.NodeName, "the name of this node")
 	fs.StringVar(&o.NodeAddress, "node-address", o.NodeAddress, "the address of this node")
+	fs.StringVar(&o.LabelSelector, "label-selector", o.LabelSelector, ""+
```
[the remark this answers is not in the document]
chenxi-seu: I miss this function, i will change it and rename to DefaultCNRLabels.

cmd/katalyst-agent/app/options/global/base.go
chenxi-seu: done

cmd/katalyst-agent/app/options/global/base.go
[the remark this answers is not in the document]
chenxi-seu: done

pkg/controller/lifecycle/cnr.go
chenxi-seu: done

waynepeking348 (dismissed):

luomingmeng (approved):

pkg/agent/resourcemanager/reporter/cnr/cnrreporter.go
```diff
@@ -61,6 +61,8 @@ const (
 type cnrReporterImpl struct {
 	cnrName string
 
+	// labels contains the default config for CNR created by reporter
+	labels map[string]string
```
caohe, line 65: Would it be better to name it defaultLabels?

chenxi-seu: yep, done

waynepeking348 (approved):

caohe (approved):

Issue #219, opened 2023-08-17
Filter the Pods reported in KCNR


zhy76:
### What would you like to be added?

The results returned by the pod resources server are not filtered, so anything returned by the pod resources server will be reported to KCNR. This may also need to be considered in conjunction with the scheduler to determine if we need to filter out only non-NUMA binding Pods.
![image](https://github.com/kubewharf/katalyst-core/assets/56665618/b87b147c-218e-4312-9ac7-7a6f5dbb77ab)

### Why is this needed?

The Pods currently reported by KCNR may need filtering

zhy76:
![image](https://github.com/kubewharf/katalyst-core/assets/56665618/c583c0de-9e7e-4e52-b70b-38438ab1416a)
This looks abnormal too

PR #256, opened 2023-09-04
feat: Filter pod resources for reclaimed qos pods and shared qos pods reported in KCNR

zhy76:
#### What type of PR is this?

Enhancements
feat: Filter the Pods reported in KCNR
Fixes: #219


waynepeking348:
why we need to filter out shared-cores?

zhy76:
According to the discussion yesterday, for the problem of kcnr exception monitoring, special treatment is taken for the monitoring logic, that is, only care about the numabinding pod, for other numabinding types of Pods that may be implemented in the future, only need to adapt the corresponding monitoring logic. So, turn off this pr first.

PR #279, opened 2023-09-14
feat(auth): auth framework

zzzzhhb:
#### What type of PR is this?
Features

This PR adds a framework for authentication and authorization which will support multiple protocol and mechanism.


pkg/agent/evictionmanager/manager.go
```diff
@@ -45,16 +45,19 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/consts"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
```
waynepeking348, line 46: for now, only eviction-manager will call the auth functions (but not for reporter .etc), right?

waynepeking348 (approved):

csfldf (approved):

PR #296, opened 2023-09-26
enhancement(eviction): numa pressure support victim minimum usage thr…

zzzzhhb:
…eshold

#### What type of PR is this?

Enhancements

This PR support NumaVictimMinimumUsageThreshold configuration. DefaultNumaVictimMinimumUsageThreshold is the victim's minimum memory usage on a NUMA node. If a pod uses less memory on a NUMA node than this threshold, it won't be evicted by this NUMA's memory pressure.


go.mod
```diff
@@ -141,6 +141,7 @@ require (
 )
 
 replace (
+	github.com/kubewharf/katalyst-api => github.com/zzzzhhb/katalyst-api v0.0.0-20230926073642-3971260393a0
```
waynepeking348, line 144: pls fix this import

pkg/metaserver/agent/metric/helper/system.go
```diff
@@ -23,78 +23,76 @@ import (
...
 
 	return free, total, scaleFactor, nil
 }
 
 func GetSystemKswapdStealMetrics(metricsFetcher metric.MetricsFetcher, emitter metrics.MetricEmitter) (metricutil.MetricData, error) {
-	kswapdSteal, err := metricsFetcher.GetNodeMetric(consts.MetricMemKswapdstealSystem)
+	return GetNodeMetricWithTime(metricsFetcher, emitter, consts.MetricMemKswapdstealSystem)
+}
+
+func GetNodeMetricWithTime(metricsFetcher metric.MetricsFetcher, emitter metrics.MetricEmitter, metricName string) (metricutil.MetricData, error) {
```
waynepeking348, line 65: `GetNodeMetricWithTime` is kind of weird, why node needs a function like this while numa not

pkg/metaserver/agent/metric/helper/system.go
```diff
@@ -23,78 +23,76 @@ import (
...
-	return kswapdSteal, nil
+func GetNodeMetric(metricsFetcher metric.MetricsFetcher, emitter metrics.MetricEmitter, metricName string) (float64, error) {
+	metricWithTime, err := GetNodeMetricWithTime(metricsFetcher, emitter, metricName)
+	if err != nil {
+		return 0, err
+	}
+	return metricWithTime.Value, err
+}
+
+func GetNumaMetric(metricsFetcher metric.MetricsFetcher, emitter metrics.MetricEmitter, metricName string, numaID int) (float64, error) {
```
waynepeking348, line 85: some functions should be exported outside while others not, we should apparently separate them, either by a different file or lowercased functions

pkg/metaserver/agent/metric/helper/system.go
zzzzhhb: It depends if there is a caller who needs timestamp of metric. For now, only node metrics faces this kind of need, so I just provide this kind of function for node metrics. I'll provide one for numa metrics to make them looks more consistent.

pkg/metaserver/agent/metric/helper/system.go
zzzzhhb: which one should not be exported outside?

waynepeking348 (approved):

cheney-lin (approved):

PR #306, opened 2023-10-08
Fix calculation of rate metrics

justadogistaken:

Due to the sampling lag between katalyst(once per 5s) and malachite-core(once per 10s), there will be lots of "zero" value produced(70% for mem_read/write_bandwidth).

This work tries to fix the problem by setting rate metric only when the metric is updated.

For the previous version:
(Many "zero" cases could be seen in the three following moments.)

截屏2023-10-09 14 57 01

截屏2023-10-09 14 57 14

截屏2023-10-09 14 57 29

After Updating:
(No more "zero" values, but at the "start" moment basically. I did't upload them all, but I had checked about it.)

截屏2023-10-09 14 28 39

### DIFF
--- a/pkg/metaserver/agent/metric/malachite/fetcher_calculate.go
+++ b/pkg/metaserver/agent/metric/malachite/fetcher_calculate.go
@@ -27,47 +27,64 @@ import (
 )
 
 // processContainerMemBandwidth handles memory bandwidth (read/write) rate in a period while,
-// and it will need the previously collected datat to do this
-func (m *MalachiteMetricsFetcher) processContainerMemBandwidth(podUID, containerName string, cgStats *types.MalachiteCgroupInfo) {
+// and it will need the previously collected data to do this
+func (m *MalachiteMetricsFetcher) processContainerMemBandwidth(podUID, containerName string, cgStats *types.MalachiteCgroupInfo, lastUpdateTimeInSec float64) {
 	var (
 		lastOCRReadDRAMs, _ = m.metricStore.GetContainerMetric(podUID, containerName, consts.MetricOCRReadDRAMsContainer)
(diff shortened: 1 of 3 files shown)


waynepeking348 (approved):

pkg/metaserver/agent/metric/malachite/fetcher_calculate.go
```diff
@@ -27,45 +27,48 @@ import (
...
+		curUpdateTimeInSec = float64(cgStats.V2.Cpu.UpdateTime)
 	}
 
 	// read/write bandwidth calculation formula
-	timediffSecs := curUpdateTime - lastUpdateTime.Value
-	if timediffSecs > 0 {
-		readBytes := (curOCRReadDRAMs - lastOCRReadDRAMs.Value) * 64
-		readBandwidth = readBytes / (1024 * 1024 * timediffSecs)
+	timeDiffSecs := curUpdateTimeInSec - lastUpdateTimeInSec
+	if timeDiffSecs <= 0 {
```
waynepeking348, line 57: @justadogistaken so this pr only fixes the case that when `timeDiffSecs` <= 0, we should skip the calculation  rather than set the values as zero by mistake. right?

justadogistaken: Yes. I found that it produced lots of "zero" value. The problem is originated from the sampling gap between katalyst(once per 5s) and malachite-core(once per 10s).

And I'm checking if there are some other metrics have the same problem, I will fix them along this pr.

waynepeking348 (dismissed):

pkg/metaserver/agent/metric/malachite/fetcher_calculate.go
waynepeking348: btw, the coding style is really elegant!

waynepeking348 (approved):

sun-yuliang (approved):

PR #321, opened 2023-10-19
feat(advisor): support disable registered model result fetcher

csfldf:
feat(advisor): support disable registered model result fetcher


justadogistaken (approved):
LGTM

waynepeking348 (approved):

PR #345, opened 2023-11-06
enhancement(sysadvisor): export region reclaimable info as non-reclaim when dedicated_cores can not be co-located with reclaimed_cores

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
@@ -17,6 +17,9 @@ limitations under the License.
 package region
 
 import (
+	"context"
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/klog/v2"
 	"k8s.io/kubelet/pkg/apis/resourceplugin/v1alpha1"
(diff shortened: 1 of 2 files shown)


sun-yuliang (dismissed):

pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/region/region_dedicated_numa_exclusive.go
```diff
@@ -38,6 +42,7 @@ import (
 
 type QoSRegionDedicatedNumaExclusive struct {
 	*QoSRegionBase
+	podUID string
```
waynepeking348, line 45: personally, I would prefer not to define podUID here, since it indicates that only contain one pod can be supported, but the podSet indicates that we should support multiple pods, and it may cause a lot of confusion as the logic goes more and more complex

waynepeking348: or we can redefine the high-level design, i.e. "whether it is reasonable to make every region to contain a podSet "

cheney-lin: It is necessary to keep the podset because we need to calculate resource usage by it. NewQoSRegionDedicatedNumaExclusive accept containerInfo as one of the parameters indicate that only this region only belong to single Pod.

waynepeking348: still confused for keeping both podSet and podUID; besides, the reason for `NewQoSRegionDedicatedNumaExclusive accept containerInfo as one of the parameters` is not a solid reason for keeping podUID or indicating there should be only one pod in this region, since all regions accept containerInfo as one of the parameters (cause we need for its owner pool name) ...

cheney-lin: well, I removed podUID of QoSRegionDedicatedNumaExclusive and use getPodUID to get it.

waynepeking348 (approved):

sun-yuliang (approved):

Issue #220, opened 2023-08-17
Support inter-pod affinity and anti-affinity at NUMA level


caohe:
### What would you like to be added?

This issue is opened to track the development of inter-pod affinity and anti-affinity at NUMA level in Kubernetes.

### Why is this needed?

Currently, Kubernetes supports inter-pod affinity and anti-affinity at the node level. However, there is a growing need for extending this support to the NUMA level. 

For example, high-memory bandwidth consuming pods, like workers, can impact the performance of other pods on the same NUMA node, such as parameter servers. Allocating these pods to different NUMA nodes can mitigate such interferences.

yanxiaoqi932:
Hello, I wanna participant it 😄

waynepeking348:
we will have community meeting to discuss this, can we post the documents here? @pendoragon 

> Hello, I wanna participant it 😄

pendoragon:
> we will have community meeting to discuss this, can we post the documents here? @pendoragon
> 
> > Hello, I wanna participant it 😄

sure, https://bytedance.feishu.cn/docx/VXGfdiddUoemoQx54uacwSHonPd @yanxiaoqi932 feel free to edit the agenda

yanxiaoqi932:
> > we will have community meeting to discuss this, can we post the documents here? @pendoragon
> > > Hello, I wanna participant it 😄
> 
> sure, https://bytedance.feishu.cn/docx/VXGfdiddUoemoQx54uacwSHonPd @yanxiaoqi932 feel free to edit the agenda

Thank you. I have applied for permission to the online document.

PR #347, opened 2023-11-07
enhancement(plugin): Support inter-pod affinity and anti-affinity at NUMA level

yanxiaoqi932:
#### What type of PR is this?
Enhancement

Support inter-pod affinity and anti-affinity at NUMA level，#220

The test results on my machine:


pkg/scheduler/cache/nodeinfo.go
```diff
@@ -129,3 +217,59 @@ func (n *NodeInfo) RemovePod(key string, pod *v1.Pod) {
...
+		return nil
+	}
+
+	return n.ResourceTopology.WithPodReousrce(n.AssumedPodResources, filterFn)
+}
+
+func (n *NodeInfo) GetAssumedPodAffinityInfo() AssumedPodAffinityInfo {
+	n.Mutex.RLock()
+	defer n.Mutex.RUnlock()
+	return n.AssumedPodAffinityInfo
```
WangZzzhe, line 248: AssumedPodAffinityInfo is not thread safe, maybe it is better to return a deepcopy ?

PR #348, opened 2023-11-08
enhancement(sysadvisor): set headroom into metaCache

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/agent/sysadvisor/metacache/metacache.go
+++ b/pkg/agent/sysadvisor/metacache/metacache.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/kubelet/checkpointmanager"
@@ -77,6 +78,8 @@ type MetaReader interface {
 	// GetPodsInferenceResult gets specified model inference result
 	GetInferenceResult(modelName string) (interface{}, error)
 
+	GetHeadroom(name types.QoSResourceName) (*resource.Quantity, error)
+
 	metric.MetricsReader
 }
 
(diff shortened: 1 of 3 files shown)


waynepeking348:
@luomingmeng pls have a look for this pr, since we can already get headroom info from meta-sever; would be better if this logic can be aligned

pkg/agent/sysadvisor/metacache/metacache.go
```diff
@@ -77,6 +78,8 @@ type MetaReader interface {
 	// GetPodsInferenceResult gets specified model inference result
 	GetInferenceResult(modelName string) (interface{}, error)
 
+	GetHeadroom(name types.QoSResourceName) (*resource.Quantity, error)
```
sun-yuliang, line 81: Not recommended to add GetHeadroom interface because it is not general. Use GetNodeInfo instead and update headroom property.

pkg/agent/sysadvisor/metacache/metacache.go
```diff
@@ -116,6 +119,8 @@ type MetaWriter interface {
 
 	// SetInferenceResult sets specified model inference result
 	SetInferenceResult(modelName string, result interface{}) error
+
+	SetHeadroom(name types.QoSResourceName, headroom resource.Quantity)
```
sun-yuliang, line 123: Same as above.

pkg/agent/sysadvisor/plugin/qosaware/resource/resource.go
```diff
@@ -62,6 +62,7 @@ type SubResourceAdvisor interface {
 
 type resourceAdvisorWrapper struct {
 	subAdvisorsToRun map[types.QoSResourceName]SubResourceAdvisor
+	metaCache        metacache.MetaCache
```
sun-yuliang, line 65: Not recommended to add metaCache here. To align with region info update, it's better to update headroom in xxxResoureAdvisor's GetHeadroom() function.

PR #365, opened 2023-11-15
unified solution for TCP memory limitation

lubinszARM:
#### What type of PR is this?
Features
The feature provides the unified solution for TCP memory limitation in cgroup and host level.
The feature includes 3 parts:
1, Set the limit value for host net.ipv4.tcp_mem. 
The default value is 20% of host toal memory.
2, Do nothing for cgroupv2.
3, Set pod tcp_mem accounting for cgroupv1. 
The default value is same with memory.limit_in_bytes.

### DIFF
--- a/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy.go
+++ b/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy.go
@@ -213,6 +213,7 @@ func NewDynamicPolicy(agentCtx *agent.GenericContext, conf *config.Configuration
 		memoryadvisor.ControlKnobHandlerWithChecker(policyImplement.handleAdvisorDropCache))
 
 	sockmem.UpdateHostTCPMemRatio(conf.SetHostTCPMemLimitRatio)
+	sockmem.UpdateCgroupTCPMemRatio(conf.SetCgroupTCPMemLimitRatio)
 
 	return true, &agent.PluginWrapper{GenericPlugin: pluginWrapper}, nil
 }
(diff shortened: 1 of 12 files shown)


pkg/agent/qrm-plugins/memory/dynamicpolicy/policy.go
```diff
@@ -245,6 +252,11 @@ func (p *DynamicPolicy) Start() (err error) {
 		go wait.Until(p.setMemoryMigrate, setMemoryMigratePeriod, p.stopCh)
 	}
 
+	if p.enableSettingSockMem {
```
waynepeking348, line 255: register as `RegisterPeriodicalHandler` will be better

pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_async_handler.go
```diff
@@ -481,3 +485,74 @@ func (p *DynamicPolicy) setMemoryMigrate() {
...
 }
+
+/*
+ * setSockMemLimit is the unified solution for tcpmem limitation.
+ * it includes 3 parts:
+ * 1, set the limit value for host net.ipv4.tcp_mem.
+ * 2, do nothing for cgroupv2.
+ * 3, set pod tcp_mem accounting for cgroupv1.
+ */
+func (p *DynamicPolicy) setSockMemLimit() {
```
waynepeking348, line 496: will it be better if we add a new dir `handlers` for `RegisterPeriodicalHandler `? @csfldf

PR #375, opened 2023-11-22
enhancement(spd): set default baselint ratio by qos level

cheney-lin:
#### What type of PR is this?

Enhancements


pkg/controller/spd/indicator-plugin/manager.go
```diff
@@ -49,10 +52,10 @@ type IndicatorGetter interface {
 	GetIndicatorSpecChan() chan types.NamespacedName
 	GetIndicatorStatusChan() chan types.NamespacedName
 
-	// GetIndicatorSpec + GetIndicatorStatus
+	// GetSPDSpec + GetSPDStatus
 	// for indicator get functions, IndicatorUpdater will return a channel to obtain the merged results.
-	GetIndicatorSpec(_ types.NamespacedName) *apiworkload.ServiceProfileDescriptorSpec
-	GetIndicatorStatus(_ types.NamespacedName) *apiworkload.ServiceProfileDescriptorStatus
+	GetSPDSpec(_ types.NamespacedName) *apiworkload.ServiceProfileDescriptorSpec
```
waynepeking348, line 57: maybe it's better to keep `GetIndicatorSpec`, since the logic is for updating indicator spec (despite that the returned value is entire spec for spd)

pkg/controller/spd/spd.go
```diff
@@ -539,6 +542,25 @@ func (sc *SPDController) getWorkload(gvr schema.GroupVersionResource, namespace,
 	return workload, nil
 }
 
+func (sc *SPDController) defaultBaselineRatio(workload *unstructured.Unstructured) *float32 {
```
waynepeking348, line 545: for safety reasons, the default value when error appears should be 100 rather than nil (which actually equals 0)

waynepeking348: and remember to add some comments for why we should return 100

pkg/controller/spd/indicator-plugin/manager.go
```diff
@@ -39,6 +40,8 @@ type IndicatorUpdater interface {
 	AddBusinessIndicatorSpec(_ types.NamespacedName, _ []apiworkload.ServiceBusinessIndicatorSpec)
 	AddSystemIndicatorSpec(_ types.NamespacedName, _ []apiworkload.ServiceSystemIndicatorSpec)
 	AddBusinessIndicatorStatus(_ types.NamespacedName, _ []apiworkload.ServiceBusinessIndicatorStatus)
+
+	UpdateBaselineRatio(_ types.NamespacedName, _ float32)
```
waynepeking348, line 44: make the naming aligned, either by change the origin `AddBusinessIndicatorSpec` to `Updatexxx` or `UpdateBaselineRatio` to `AddBaselineRatio`

waynepeking348 (approved):

luomingmeng (approved):

PR #388, opened 2023-11-30
 fix(spd): parseBaselineCoefficient returns error if BaselineCoefficient is empty

cheney-lin:
#### What type of PR is this?

Bug fixes

    If baselinePercent is set as 100%, annotation.SPDAnnotationBaselinePercentileKey will
    be updated as empty, error occurs when parseBaselineCoefficient, which prints as followed:
    "[update] [qosaware-memory] update headroom policy failed" err="index 0 cannot parse: strconv.Atoi: parsing \"\": invalid syntax" headroomPolicy=numa-aware
    
    By the way, I refined relative logic, removing confuesd BaselinePercentile value and replace by human readable sentinel pod.


waynepeking348 (approved):

pkg/util/spd_baseline.go
```diff
@@ -17,46 +17,44 @@ limitations under the License.
...
 	"github.com/kubewharf/katalyst-api/pkg/apis/workload/v1alpha1"
 	"github.com/kubewharf/katalyst-api/pkg/consts"
 )
 
-const (
-	baselineCoefficientSep = ","
-)
-
-type BaselineCoefficient []int64
+type PodMeta struct {
```
luomingmeng, line 30: To avoid conflict with other definitions, consider adding a prefix of spd baseline

cheney-lin: modified

waynepeking348 (approved):

PR #420, opened 2023-12-22
enhancement(eviction): refine memory pressure eviction

LuyaoZhong:
Add configurations to enable/disable eviction when detecting free below watermark at first time. Add watermark check when detecting kswapd steal pressure to avoid eviction on memory fragmentation.


pkg/agent/evictionmanager/plugin/memory/enhanced_numa_pressure.go
```diff
@@ -0,0 +1,305 @@
...
+				general.Infof("Failed to get pod  %+v total mem usage on numa id %+v", pod.UID, numaID)
+				continue
+			}
+			// TODO: it's better to use sum of podNumaAnon and podNumaActiveFile to predict how much the pod can release the pressure
+			//  since inactive file is already counted in predictNumaPressureStatMap, while cgroupv1 does not provide numa active/inactive statistics.
+			//  So we falls back to use pod numa total.
+			predictNumaPressureStatMap[numaID] += podNumaTotal
+		}
+	}
+	getPodPermutations := func(podList []*v1.Pod) [][]*v1.Pod {
```
zzzzhhb, line 230: when `podList` get larger, this function will be costly, is it really necessary?

pkg/metaserver/agent/metric/provisioner/malachite/provisioner.go
cheney-lin, line 1: The unit of NumaStatsV1 is page, while the unit of NumaStatsV2 is byte, so there is no need to modify it here.

PR #423, opened 2023-12-27
add model ouput to kcmas

justadogistaken:
Emitting model inference result of pod to kcmas. Which could help centric components leverage the output to work on optimization stuff.

### DIFF
--- a/pkg/agent/sysadvisor/plugin/inference/inference.go
+++ b/pkg/agent/sysadvisor/plugin/inference/inference.go
@@ -29,6 +29,8 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/agent/sysadvisor/plugin/inference/modelresultfetcher"
 	borweinfetcher "github.com/kubewharf/katalyst-core/pkg/agent/sysadvisor/plugin/inference/modelresultfetcher/borwein"
 	"github.com/kubewharf/katalyst-core/pkg/config"
+	metricemitter "github.com/kubewharf/katalyst-core/pkg/config/agent/sysadvisor/metric-emitter"
+	"github.com/kubewharf/katalyst-core/pkg/config/generic"
 	"github.com/kubewharf/katalyst-core/pkg/metaserver"
 	"github.com/kubewharf/katalyst-core/pkg/metrics"
 	metricspool "github.com/kubewharf/katalyst-core/pkg/metrics/metrics-pool"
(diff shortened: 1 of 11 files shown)


pkg/agent/sysadvisor/plugin/inference/inference.go
```diff
@@ -127,3 +163,106 @@ func (infp *InferencePlugin) fetchModelResult(ctx context.Context) {
 
 	wg.Wait()
 }
+
+func (infp *InferencePlugin) syncModelResults(ctx context.Context) {
```
waynepeking348, line 167: not suggest to emit kcmas-related metrics everywhere, maybe it's better to put in metric package instead

justadogistaken: I agree with you, but why I emit kcmas-metrics is it's hard to reuse podSyncer under metric package.
So do you accept that is goes as:
Create a new file under metric package. And define an exported function(not belongs to any struct). Then call it under `InferencePlugin.Run`?

sun-yuliang (approved):

pkg/agent/sysadvisor/plugin/inference/models/borwein/types/types.go
```diff
@@ -33,4 +33,37 @@ type BorweinParameter struct {
 // the first key is the pod UID
```
csfldf, line 33: move key description to Results property?

pkg/agent/sysadvisor/plugin/inference/inference.go
```diff
@@ -68,14 +74,25 @@ func NewInferencePlugin(pluginName string, conf *config.Configuration, extraConf
 		return nil, fmt.Errorf("nil emitterPool")
 	}
 
-	emitter := emitterPool.GetDefaultMetricsEmitter().WithTags("advisor-inference")
+	metricEmitter := emitterPool.GetDefaultMetricsEmitter().WithTags("advisor-inference")
+
+	dataEmitter, err := emitterPool.GetMetricsEmitter(metricspool.PrometheusMetricOptions{
```
csfldf, line 79: it seems that dataEmitter isn't used in inferencePlugin?

justadogistaken: forgot to delete that.
removed already.

sun-yuliang (approved):

nightmeng (approved):

luomingmeng (approved):

zzzzhhb (approved):

csfldf (approved):

waynepeking348 (approved):

PR #429, opened 2024-01-02
feat(sysadvisor): support numa memory balance

zzzzhhb:
#### What type of PR is this?
Features

This PR support to relief hotspot NUMA memory pressure by migrating memory between NUMAs when there is a NUMA hotspot.

### DIFF
--- a/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
+++ b/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
@@ -534,35 +534,146 @@ func (p *DynamicPolicy) handleNumaMemoryBalance(_ *config.Configuration,
 			memoryadvisor.ControlKnobKeyBalanceNumaMemory, value, err)
 	}
 
-	containerID, err := p.metaServer.GetContainerID(entryName, subEntryName)
-	if err != nil {
-		general.Errorf("get container id of pod: %s container: %s failed with error: %v", entryName, subEntryName, err)
-		return err
+	//containerID, err := p.metaServer.GetContainerID(entryName, subEntryName)
+	//if err != nil {
+	//	general.Errorf("get container id of pod: %s container: %s failed with error: %v", entryName, subEntryName, err)
+	//	return err
+	//}
+	//
+	//sourceNumaSet := machine.NewCPUSet(advice.SourceNuma)
+	//destNumaSet := machine.NewCPUSet(advice.DestNumas...)
+
+	for poolName, poolAdvice := range advice.PoolBalanceInfo {
+		migratePagesWorkName := fmt.Sprintf("%v-%s", memoryadvisor.ControlKnobKeyBalanceNumaMemory, poolName)
+		// start a asynchronous work to migrate pages for containers
+		err = p.asyncWorkers.AddWork(migratePagesWorkName,
+			&asyncworker.Work{
+				Fn:          p.doNumaMemoryBalance,
+				Params:      []interface{}{poolName, poolAdvice},
+				DeliveredAt: time.Now()})
+
+		if err != nil {
+			general.Errorf("add work: %s for pool %s failed with error: %v", migratePagesWorkName, poolName, err)
+		}
+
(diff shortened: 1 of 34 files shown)


waynepeking348 (approved):

pkg/util/cgroup/manager/v1/fs_linux.go
```diff
@@ -231,6 +232,42 @@ func (m *manager) GetMemory(absCgroupPath string) (*common.MemoryStats, error) {
 	return memoryStats, nil
 }
 
+func (m *manager) GetNumaMemory(absCgroupPath string) (map[int]*common.MemoryNumaMetrics, error) {
```
waynepeking348, line 235: will it be better if we move all the `get` function to metric-fetcher as a new implementation?

pkg/util/asyncworker/async_workers.go
```diff
@@ -259,3 +259,20 @@ func (aws *AsyncWorkers) cleanupWorkStatus() {
 		}
 	}
 }
+
+func (aws *AsyncWorkers) WorkExists(workName string) bool {
```
waynepeking348, line 263: it seems `AddWork` has already handled the duplicate logic, we don't need to have this judgement

pkg/util/cgroup/manager/v1/fs_linux.go
zzzzhhb: it's hard to say, these `get` functions only cover a little part of all metric-fetcher scenario so if we do it there will be a lot of effort. Anyway, we shouldn't do it in this PR, maybe we can discuss it later then create a new PR to do that.

pkg/util/asyncworker/async_workers.go
zzzzhhb: there's a little difference, the `AddWork` function will stop current work and launch the newer work but in this scenario I just want to discard the newer work when there's a running work.

waynepeking348: prefer to add a parameter in `AddWork` rather than do it in caller function; not stick to it though, @csfldf what do you think?

pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
```diff
@@ -516,3 +517,165 @@ func (p *DynamicPolicy) handleAdvisorMemoryProvisions(_ *config.Configuration,
...
+type containerMigrateStat struct {
+	containerID string
+	cgroupPath  string
+	rssBefore   uint64
+	rssAfter    uint64
+}
+
+func (p *DynamicPolicy) doNumaMemoryBalance(ctx context.Context, advice types.NumaMemoryBalanceAdvice) error {
+	startTime := time.Now()
+	defer func() {
```
csfldf, line 568: there is time cost [metrics](https://github.com/kubewharf/katalyst-core/blob/a5b8bd30253a35e8842712335fde42ae75be6668/pkg/util/asyncworker/async_workers.go#L165) in the async worker framework, do we still need logic here?

pkg/util/asyncworker/async_workers.go
csfldf: I agree with adding a parameter in AddWork instead of judging it by the user

zzzzhhb: get, I'll try to refine it.

pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
zzzzhhb: if so, we don't need this anymore. I'll remove this logic.

waynepeking348 (dismissed):

sun-yuliang (dismissed):

csfldf (dismissed):

waynepeking348 (approved):

csfldf (approved):

sun-yuliang (approved):

luomingmeng (approved):

nightmeng (approved):

PR #435, opened 2024-01-08
feat(ORM): outofband topology manager

WangZzzhe:
#### What type of PR is this?
Enhancements

Add topology manager with policies(none, best-effort, restrict, single-numa-node, numeric) for ORM.
Pods added by ORM should GetTopologyHints from QRM plugins and generated a best hint by topology manager policy before allocate resource. Dedicated-cores pods with numaBinding can be allocated resource by this way in ORM.

### DIFF
--- a/pkg/agent/orm/topology/manager.go
+++ b/pkg/agent/orm/topology/manager.go
@@ -1,0 +1,224 @@
+/*
+Copyright 2022 The Katalyst Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package topology
+
+import (
+	"fmt"
+	"sync"
+
+	cadvisorapi "github.com/google/cadvisor/info/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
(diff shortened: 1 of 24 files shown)


waynepeking348 (approved):

waynepeking348 (approved):

waynepeking348:
@caohe pls have a look for this pr

pkg/agent/orm/topology/manager.go
```diff
@@ -0,0 +1,224 @@
...
+// NUMA locality.
+type HintProvider interface {
+	// GetTopologyHints returns a map of resource names to a list of possible
+	// concrete resource allocations in terms of NUMA locality hints. Each hint
+	// is optionally marked "preferred" and indicates the set of NUMA nodes
+	// involved in the hypothetical allocation. The topology manager calls
+	// this function for each hint provider, and merges the hints to produce
+	// a consensus "best" hint. The hint providers may subsequently query the
+	// topology manager to influence actual resource assignment.
+	GetTopologyHints(pod *v1.Pod, container *v1.Container) map[string][]TopologyHint
```
caohe, line 64: Do we also need `GetPodTopologyHints`?

WangZzzhe: > Do we also need `GetPodTopologyHints`?

yes,  if we want to allocate resources for dedicated cores pods on specific NUMA nodes, this is the only way to get available hints from QRMPlugins.

caohe: It seems `GetPodTopologyHints` has not been defined and implemented yet. Could you also add it?

WangZzzhe: > It seems `GetPodTopologyHints` has not been defined and implemented yet. Could you also add it?

sorry I misunderstood you question, `GetPodTopologyHints` is not implemented by QRM plugins so we don't need it. 
I add `GetPodTopologyHints` for ORM manager and it only returns nil now.

caohe (approved):

waynepeking348 (approved):

csfldf (approved):

PR #436, opened 2024-01-09
enhancement: support disk type and  io-latency cgroup operations

lubinszARM:
#### What type of PR is this?
Enhancements

For io qos features, when we deal with dirtymemory, we need to know the disk type. 
When we set io.latency, we need io.latency related cgroup operations.

None.

None.

### DIFF
--- a/pkg/metaserver/agent/metric/helper/device.go
+++ b/pkg/metaserver/agent/metric/helper/device.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2022 The Katalyst Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helper
+
+import (
+	"github.com/kubewharf/katalyst-core/pkg/metaserver/agent/metric/types"
+	"github.com/kubewharf/katalyst-core/pkg/metrics"
+	"github.com/kubewharf/katalyst-core/pkg/util/general"
+	metricutil "github.com/kubewharf/katalyst-core/pkg/util/metric"
+)
+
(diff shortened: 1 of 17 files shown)


waynepeking348:
this pr is based on a out-of-dated version, can't be reviewed at all @lubinszARM

waynepeking348:
also, pls add labels

pkg/util/cgroup/manager/v2/fs_linux.go
```diff
@@ -219,6 +219,34 @@ func (m *manager) ApplyIOWeight(absCgroupPath string, devID string, weight uint6
...
 
+func (m *manager) ApplyIOLatency(absCgroupPath string, devID string, latency uint64) error {
+	var dataContent string
+	if latency == 0 {
+		dataContent = fmt.Sprintf("%s target=max", devID)
+	} else {
+		dataContent = fmt.Sprintf("%s target=%d", devID, latency)
+	}
+
+	curLatency, found, err := m.GetDeviceIOLatency(absCgroupPath, devID)
```
waynepeking348, line 230: pls remove those unnecessary blanks (not only for this specific one but for all newly added codes)

pkg/metaserver/agent/metric/helper/device.go
```diff
@@ -0,0 +1,46 @@
...
+package helper
+
+import (
+	"github.com/kubewharf/katalyst-core/pkg/metaserver/agent/metric/types"
+	"github.com/kubewharf/katalyst-core/pkg/metrics"
+	"github.com/kubewharf/katalyst-core/pkg/util/general"
+	metricutil "github.com/kubewharf/katalyst-core/pkg/util/metric"
+)
+
+func GetDeviceMetricWithTime(metricsFetcher types.MetricsFetcher, emitter metrics.MetricEmitter, metricName string, devName string) (metricutil.MetricData, error) {
```
waynepeking348, line 26: we need make an appointment here (it may be a little confusing now): for a certain metric, when should we put the get-logic in `helper` package? in the previous design,  `helper` package will only be used when we want to assemble or re-calculate based on the raw-metric, but it seems not for now; and we'd better add comments in helper.consts.go @nightmeng

pkg/util/cgroup/manager/v2/fs_linux.go
```diff
@@ -219,6 +219,34 @@ func (m *manager) ApplyIOWeight(absCgroupPath string, devID string, weight uint6
 	return nil
 }
 
+func (m *manager) ApplyIOLatency(absCgroupPath string, devID string, latency uint64) error {
```
waynepeking348, line 222: will this function only be called out-of-tree (and why)?  cc @csfldf

PR #445, opened 2024-01-15
refine(qrm): calculate pool size with much accuracy

csfldf:
refine(qrm): calculate pool size with much accuracy

### DIFF
--- a/pkg/agent/qrm-plugins/cpu/dynamicpolicy/policy_allocation_handlers.go
+++ b/pkg/agent/qrm-plugins/cpu/dynamicpolicy/policy_allocation_handlers.go
@@ -19,6 +19,8 @@ package dynamicpolicy
 import (
 	"context"
 	"fmt"
+	"math"
+	"sort"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -43,7 +45,7 @@ func (p *DynamicPolicy) sharedCoresAllocationHandler(_ context.Context,
 		return nil, fmt.Errorf("sharedCoresAllocationHandler got nil request")
 	}
 
-	reqInt, err := util.GetQuantityFromResourceReq(req)
+	_, reqFloat64, err := util.GetQuantityFromResourceReq(req)
 	if err != nil {
 		return nil, fmt.Errorf("getReqQuantityFromResourceReq failed with error: %v", err)
 	}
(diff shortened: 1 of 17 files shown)


pkg/agent/qrm-plugins/cpu/dynamicpolicy/policy_allocation_handlers.go
```diff
@@ -911,20 +922,37 @@ func (p *DynamicPolicy) generatePoolsAndIsolation(poolsQuantityMap map[string]in
 		proportionalPoolsQuantityMap := make(map[string]int)
 
 		for poolName, poolQuantity := range poolsQuantityMap {
```
waynepeking348, line 924: can those logic be refined (for instance, put in another function)? kind of too large to understand ...

pkg/agent/qrm-plugins/cpu/dynamicpolicy/state/util.go
```diff
@@ -146,10 +149,16 @@ func GetSharedQuantityMapFromPodEntries(podEntries PodEntries, ignoreAllocationI
 			}
 
 			if poolName := allocationInfo.GetOwnerPoolName(); poolName != advisorapi.EmptyOwnerPoolName {
-				ret[poolName] += GetContainerRequestedCores()(allocationInfo)
+				preciseQuantityMap[poolName] += GetContainerRequestedCores()(allocationInfo)
 			}
 		}
 	}
+
+	for poolName, preciseQuantity := range preciseQuantityMap {
```
waynepeking348, line 157: will it exceeds node total if all pools are ceiled?

csfldf: yes, the sum of ret will exceed node total here. but the generatePoolsAndIsolation will adjust to guarantee cpuset of every pool won't exceed current available cpus.

pkg/agent/qrm-plugins/cpu/dynamicpolicy/policy_allocation_handlers.go
```diff
@@ -911,20 +922,37 @@ func (p *DynamicPolicy) generatePoolsAndIsolation(poolsQuantityMap map[string]in
...
-			proportionalSize := general.Max(getProportionalSize(poolQuantity, poolsTotalQuantity, availableSize), 1)
+			proportionalSize := general.Max(getProportionalSize(poolQuantity, poolsTotalQuantity, availableSize, true /*ceil*/), 1)
```
luomingmeng, line 925: the comment `ceil` means `true` will ceil the result, and the `false` is not?

csfldf: yes

waynepeking348 (approved):

luomingmeng (approved):

PR #458, opened 2024-01-24
feat(async_work): add async work duration and panic metrics

csfldf:
feat(async_work): add async work duration and panic metrics

### DIFF
--- a/pkg/util/asyncworker/async_workers.go
+++ b/pkg/util/asyncworker/async_workers.go
@@ -116,6 +116,17 @@ func (aws *AsyncWorkers) handleWork(ctx context.Context, workName string, work *
 	defer func() {
 		if r := recover(); r != nil {
 			handleErr = fmt.Errorf("recover from %v", r)
+
+			metricErr := EmitCustomizedAsyncedMetrics(ctx,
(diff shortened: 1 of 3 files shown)


pkg/util/asyncworker/async_workers.go
```diff
@@ -116,6 +116,17 @@ func (aws *AsyncWorkers) handleWork(ctx context.Context, workName string, work *
 	defer func() {
 		if r := recover(); r != nil {
 			handleErr = fmt.Errorf("recover from %v", r)
+
+			metricErr := EmitCustomizedAsyncedMetrics(ctx,
```
waynepeking348, line 120: will workName always be a valid metric-name? will it contain some non-supported characters?

csfldf: In addition to existing EmitAsyncedMetrics, I add another function EmitCustomizedAsyncedMetrics in this pr. And it accept metric name as parameter. I use specified metric "name async_work_duration_ms" and "async_work_panic" in this pr and they are both valid names.

zzzzhhb (approved):

PR #469, opened 2024-01-31
feat(metric): add rodan metric provisioner

WangZzzhe:
#### What type of PR is this?
Enhancements

Add a metric provisioner that requests metrics from rodan(metrics collector like malachite).

Add a config which allows metric fetcher to init provisioners according to the provisioner name list. Malachite and kubelet are used by default, and rodan is used in private scenarios.

### DIFF
--- a/cmd/katalyst-agent/app/options/metaserver/metaserver.go
+++ b/cmd/katalyst-agent/app/options/metaserver/metaserver.go
@@ -56,6 +56,8 @@ const defaultCustomNodeResourceCacheTTL = 15 * time.Second
 
 const defaultCustomNodeConfigCacheTTL = 15 * time.Second
 
+const defaultRodanServerPort = 9102
+
 // MetaServerOptions holds all the configurations for metaserver.
 // we will not try to separate this structure into several individual
 // structures since it will not be used directly by other components; instead,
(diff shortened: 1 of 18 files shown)


waynepeking348:
this pr should also be rebased

cmd/katalyst-agent/app/options/metaserver/metaserver.go
```diff
@@ -150,6 +154,8 @@ func (o *MetaServerOptions) AddFlags(fss *cliflag.NamedFlagSets) {
 		"The meta server return metric data and MetricDataExpired if the update time of metric data is earlier than this period.")
 	fs.StringSliceVar(&o.MetricProvisions, "metric-provisioners", o.MetricProvisions,
 		"The provisioners that should be enabled by default")
+	fs.IntVar(&o.RodanServerPort, "rodan-server-port", o.RodanServerPort,
```
caohe, line 157: Could rodan and malachite use the same config?

WangZzzhe: > Could rodan and malachite use the same config?

It would be better for each provisioner to use separate configurations, because all provisioners can be enabled simultaneously by 'metric-provisioners'.

caohe (dismissed):

caohe (approved):

nightmeng (approved):

PR #470, opened 2024-01-31
perf(qrm): improve performance of calculating hints

csfldf:
perf(qrm): improve performance of calculating hints


pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_test.go
```diff
@@ -3418,3 +3427,132 @@ func TestDynamicPolicy_adjustAllocationEntries(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkGetTopologyHints(b *testing.B) {
+	klog.SetOutput(ioutil.Discard)
+	klog.V(0)
+	klog.LogToStderr(false)
+	topoCases := []topoTestCase{
+		//{
```
waynepeking348, line 3436: maybe it's better not to keep those commented codes

cloudaice: use b.ResetTimer() before real benchmark code

PR #473, opened 2024-02-07
kubelet reporter plugin support shared_cores with numa_binding pod

luomingmeng:
#### What type of PR is this?

Enhancements
kubelet reporter plugin support report shared_cores with numa_binding pod, and add GenericPodResourcesFilter to filter out cpu and memory of shared_cores without numa_binding pod.

### DIFF
--- a/pkg/agent/resourcemanager/fetcher/kubelet/topology/topology_adapter.go
+++ b/pkg/agent/resourcemanager/fetcher/kubelet/topology/topology_adapter.go
@@ -33,11 +33,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	podresv1 "k8s.io/kubelet/pkg/apis/podresources/v1"
+	resourceutil "k8s.io/kubernetes/pkg/api/v1/resource"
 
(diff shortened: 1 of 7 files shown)


pkg/agent/resourcemanager/fetcher/kubelet/topology/topology_adapter.go
```diff
@@ -488,6 +502,50 @@ func (p *topologyAdapterImpl) getZoneAllocations(podList []*v1.Pod, podResources
...
+}
+
+// reviseSharedCoresPodAllocated is to revise shared_cores pod allocated according to its numa binding
+func (p *topologyAdapterImpl) reviseSharedCoresPodAllocated(pod *v1.Pod, podAllocated map[util.ZoneNode]*v1.ResourceList) error {
+	if !qos.IsPodNumaBinding(p.qosConf, pod) {
+		return nil
+	}
+
+	var bindingNumaNode *util.ZoneNode
+	for zoneNode, resourceList := range podAllocated {
```
waynepeking348, line 528: move this judgment to util since we may need to judge whether a pod is only for single-numa multiple times during execution

PR #475, opened 2024-02-19
chore(agent): monitor agent start

zzzzhhb:
#### What type of PR is this?
Enhancements

This PR emits a metric when agent starts.

### DIFF
--- a/cmd/katalyst-agent/app/agent.go
+++ b/cmd/katalyst-agent/app/agent.go
@@ -38,7 +38,10 @@ import (
 
 const healthzNameLockingFileAcquired = "LockingFileReady"
 
-const metricsNameLockingFailed = "get_lock_failed"
+const (
+	metricsNameLockingFailed = "get_lock_failed"
+	metricsNameAgentStarted  = "agent_started"
+)
 
 // Run starts common and uniformed agent components here, and starts other
 // specific components in other separate repos (with common components as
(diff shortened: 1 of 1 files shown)


cmd/katalyst-agent/app/agent.go
```diff
@@ -159,7 +167,7 @@ func acquireLock(genericCtx *agent.GenericContext, conf *config.Configuration) *
 	for {
 		lock, err := general.GetUniqueLock(conf.LockFileName)
 		if err != nil {
-			_ = genericCtx.EmitterPool.GetDefaultMetricsEmitter().StoreInt64(metricsNameLockingFailed, 1, metrics.MetricTypeNameRaw)
+			_ = genericCtx.EmitterPool.GetDefaultMetricsEmitter().StoreInt64(metricsNameLockingFailed, 1, metrics.MetricTypeNameCount)
```
csfldf, line 170: why use count type instead of raw type here?

zzzzhhb: because the raw type always report the last value, so for a certain instance it always report the value 1 no matter how many times it restarts. Count type can reflect the frequency. In this scenario, we should care about the restart frequency.

waynepeking348: should we add agent-component as tag here?

zzzzhhb: do you mean the  agent-component like sysadvisor or qrm-plugins in bytedance environment?

zzzzhhb: @csfldf sorry. I just noticed I changed this line by mistake, I thought this is the new metric. I'll revert this change.

waynepeking348: > do you mean the agent-component like sysadvisor or qrm-plugins in bytedance environment?

yeah, otherwise how will you distinguish between different component?

zzzzhhb: in bytedance environment, we have psm. For open source users, it seems like we have no concept of agent component for now. Users can combine any agents they want together?

waynepeking348 (approved):

PR #490, opened 2024-02-29
customized plugin manager to skipping running if no handles and registered

waynepeking348:
#### What type of PR is this?
bug fixes

customized plugin manager to skipping running if no handles and registered to avoid logging error logs too frequently

### DIFF
--- a/cmd/katalyst-agent/app/agent/context.go
+++ b/cmd/katalyst-agent/app/agent/context.go
@@ -20,12 +20,15 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
+	"go.uber.org/atomic"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/kubelet/config"
 	"k8s.io/kubernetes/pkg/kubelet/pluginmanager"
+	"k8s.io/kubernetes/pkg/kubelet/pluginmanager/cache"
 
 	"github.com/kubewharf/katalyst-api/pkg/plugins/skeleton"
 	katalystbase "github.com/kubewharf/katalyst-core/cmd/base"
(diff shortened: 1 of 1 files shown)


cmd/katalyst-agent/app/agent/context.go
```diff
@@ -109,6 +111,31 @@ func (c *GenericContext) Run(ctx context.Context) {
...
+// customizedPluginManager requires that handlers must be added before
+// it really runs to avoid logging error logs too frequently
+type customizedPluginManager struct {
+	enabled *atomic.Bool
+	pluginmanager.PluginManager
+}
+
+// Run successes iff at-least-one handler has been registered
+func (p *customizedPluginManager) Run(sourcesReady config.SourcesReady, stopCh <-chan struct{}) {
+	if p.enabled == nil || !p.enabled.Load() {
```
csfldf, line 123: should we deal with the condition that PluginManager.Run is called before PluginManager.AddHandler? 

It's similar to [shared_informer adds listener after it's started](https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L611).

cmd/katalyst-agent/app/agent/context.go
```diff
@@ -109,6 +111,31 @@ func (c *GenericContext) Run(ctx context.Context) {
...
 }
 
+// customizedPluginManager requires that handlers must be added before
+// it really runs to avoid logging error logs too frequently
+type customizedPluginManager struct {
+	enabled *atomic.Bool
+	pluginmanager.PluginManager
+}
+
+// Run successes iff at-least-one handler has been registered
```
zzzzhhb, line 121: "iff" typo?

cmd/katalyst-agent/app/agent/context.go
luomingmeng: in our case, the AddHandler must be called before Run, which is guaranteed by the agent framework.

luomingmeng (approved):

cmd/katalyst-agent/app/agent/context.go
waynepeking348: actually, iff = if and only if

csfldf (dismissed):

cmd/katalyst-agent/app/agent/context.go
```diff
@@ -109,6 +112,30 @@ func (c *GenericContext) Run(ctx context.Context) {
...
+	enabled atomic.Bool
+	pluginmanager.PluginManager
+}
+
+// Run successes iff at-least-one handler has been registered
+func (p *customizedPluginManager) Run(sourcesReady config.SourcesReady, stopCh <-chan struct{}) {
+	for range time.Tick(time.Second) {
+		if p.enabled.Load() {
+			p.PluginManager.Run(sourcesReady, stopCh)
+			klog.Infof("started plugin-manager since handlers have been enabled")
```
zzzzhhb, line 127: this log seems meaningless since the `PluginManager.Run` function only returns when it stops

luomingmeng (approved):

csfldf (approved):

zzzzhhb (approved):

PR #495, opened 2024-03-04
spd support extended indicator sdk and extended baseline

luomingmeng:
#### What type of PR is this?

Enhancements


pkg/metaserver/spd/manager.go
```diff
@@ -106,6 +117,50 @@ type serviceProfilingManager struct {
 	fetcher SPDFetcher
 }
 
+func (m *serviceProfilingManager) ServiceExtendedIndicator(ctx context.Context, pod *v1.Pod, indicators interface{}) (bool, error) {
```
LuyaoZhong, line 120: can we return indicator object as well? because we might need to fetch configs from extended indicator.

waynepeking348 (approved):

PR #497, opened 2024-03-04
feat(meatserver): use multiple subsystems to detect cgroup path

zzzzhhb:
#### What type of PR is this?
Bug fixes

This PR uses multiple cgroup subsystem to detect real cgroup path in case one of them does't exist.

### DIFF
--- a/pkg/util/cgroup/common/types.go
+++ b/pkg/util/cgroup/common/types.go
@@ -45,6 +45,9 @@ const (
 	SystemdRootPathBurstable  = "/kubepods.slice/kubepods-burstable.slice"
 )
 
+// defaultSelectedSubsysList cgroupv1 most common subsystems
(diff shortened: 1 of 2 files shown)


csfldf (dismissed):

pkg/util/cgroup/common/types.go
```diff
@@ -45,6 +45,9 @@ const (
 	SystemdRootPathBurstable  = "/kubepods.slice/kubepods-burstable.slice"
 )
 
+// DefaultSelectedSubsys cgroupv1 default select subsys
+var defaultSelectedSubsysList = []string{"cpu", "memory", "cpuset"}
```
csfldf, line 49: use constant values above? 

CgroupSubsysCPU, CgroupSubsysMemory, CgroupSubsysCPUSet

zzzzhhb: get

waynepeking348 (approved):

csfldf (approved):

PR #498, opened 2024-03-04
feat(qrm): support shared_cores with numa_binding

csfldf:
1. feat(qrm): memory plugin support shared_cores with numa_binding
2. feat(qrm): cpu plugin support shared_cores with numa_binding


pkg/agent/qrm-plugins/cpu/dynamicpolicy/state/state.go
```diff
@@ -391,6 +574,23 @@ func (ns *NUMANodeState) ExistMatchedAllocationInfo(f func(ai *AllocationInfo) b
 	return false
 }
 
+// ExistMatchedAllocationInfo returns true if the stated predicate (with annotations of candidate)
```
waynepeking348, line 577: ExistMatchedAllocationInfo -> ExistMatchedAllocationInfoWithAnnotations

Issue #492, opened 2024-02-29
Promote node resource over-commitment to GA


caohe:
#### Why is this needed?
In v0.4, we released the MVP version of node resource over-commitment and implemented some basic features.

In v0.5, we plan to make some enhancements to this function to bring it to GA status.

#### What would you like to be added?
- [ ] **Dynamic over-commitment ratio adjustment**: In order to make the amount of over-committed resources more accurate, we will combine **long-term** and **short-term** prediction algorithms to calculate the amount of resources that can be over-committed.  https://github.com/kubewharf/katalyst-core/pull/472
- [ ] **Interference detection and mitigation**: In order to avoid resource competition caused by over-commitment, we will introduce multi-dimensional interference detection strategies, including CPU load/usage, memory usage, the reclaiming rate of kswapd, etc. Furthermore, we will introduce multi-tiered mitigation measures, including scheduling prevention, eviction, etc.  https://github.com/kubewharf/katalyst-core/pull/518
- [ ] **Compatibility with core binding**: Prevent the bound cores from being over-committed to avoid scheduling too many CPU-bound Pods and causing the Pods to fail to start.  https://github.com/kubewharf/katalyst-core/pull/472

PR #518, opened 2024-03-21
feat(eviction): add cpu system pressure eviction plugin

WangZzzhe:
#### What type of PR is this?
Enhancements

- add cpu system pressure eviction plugin, which collects cpu usage and load and verifies if these metrics are overused. It will be enabled in the scenarios where the node does not enable co-location. 
- add k8s native QoS sorter and pod owner sorter, which can be enabled through configuration in memory system eviction plugin.
- MetricRing has been moved to util.metric  

https://github.com/kubewharf/katalyst-core/issues/492
PR in katalyst-api: https://github.com/kubewharf/katalyst-api/pull/74     go.mod should be updated after api PR merged


waynepeking348:
1. what's the difference between this eviction plugin and cpu-load && cpu-suppression eviction plugins? 
2. why don't we implement this eviction in qrm but in eviction-manager? will it be conflict with qrm hybrid logic?

WangZzzhe:
> 1. what's the difference between this eviction plugin and cpu-load && cpu-suppression eviction plugins?
> 2. why don't we implement this eviction in qrm but in eviction-manager? will it be conflict with qrm hybrid logic?

1. This eviction plugin is supposed to be used in non-colocation scenarios(eg. overcommit), detecting the load of node rather than resource pool and selecting evicted pods by native QoS.
2. Qrm plugins will not be run when colocation is not applied in user's clusters, only eviction plugins will run based on metaServer. 
If colocation is applied, cpu-load and cpu-suppression will be used and this plugin should be disabled.   @waynepeking348

pkg/agent/evictionmanager/plugin/cpu/system_pressure.go
```diff
@@ -0,0 +1,554 @@
...
+		topN          = general.MinUInt64(request.TopN, uint64(len(candidatePods)))
+		now           = time.Now()
+		dynamicConfig = s.dynamicConf.GetDynamicConfiguration()
+		topNPods      = make([]*v1.Pod, 0)
+	)
+
+	s.Lock()
+	defer s.Unlock()
+
+	if !dynamicConfig.EnableCPUSystemEviction {
```
zzzzhhb, line 292: why not check this configuration immediately after entering this function

pkg/agent/evictionmanager/plugin/cpu/system_pressure.go
```diff
@@ -0,0 +1,554 @@
...
+		klog.Warningf("EnableCPUSystemEviction off, return")
+		return &v1alpha1.GetTopEvictionPodsResponse{}, nil
+	}
+	if s.overMetricName == "" {
+		klog.Warningf("cpu system eviction without over metric name, skip")
```
zzzzhhb, line 297: same question as above

cmd/katalyst-agent/app/options/dynamic/adminqos/eviction/cpu_system_eviction.go
```diff
@@ -0,0 +1,146 @@
...
+	defaultSystemUsageUpperBoundRatio           = 0.8
+	defaultSystemUsageLowerBoundRatio           = 0.6
+	defaultThresholdMetPercentage               = 0.8
+	defaultMetricRingSize                       = 10
+	defaultEvictionCoolDownTime                 = 300 * time.Second
+	defaultCheckCPUManager                      = false
+	defaultCPUSystemPressureEvictionGracePeriod = -1
+)
+
+var defaultEvictionRankingMetrics = []string{"qos.pod", "native.qos.pod", "priority.pod"}
```
zzzzhhb, line 41: why not use  const

zzzzhhb (approved):

PR #558, opened 2024-04-23
feat: agent get SPD syncly

ddjjia:
#### What type of PR is this?
Features

get spd from remote if not in cache when GetFromRemote is true


pkg/metaserver/spd/fetcher.go
```diff
@@ -159,6 +161,28 @@ func (s *spdFetcher) getSPDByNamespaceName(_ context.Context, namespace, name st
 	currentSPD := s.spdCache.GetSPD(key, true)
 	if currentSPD != nil {
 		return currentSPD, nil
+	} else if s.getFromRemote {
+		klog.Infof("[spd-manager] need to get spd %s from remote", key)
```
luomingmeng, line 165: this logic in the getSPDByNamespaceName can be abstracted into a function with the logic in sync.

PR #561, opened 2024-04-26
feat(controller): spd creation improvement

ddjjia:


pkg/controller/spd/indicator-plugin/manager.go
```diff
@@ -39,7 +40,7 @@ type IndicatorUpdater interface {
 	UpdateExtendedIndicatorSpec(_ types.NamespacedName, _ []apiworkload.ServiceExtendedIndicatorSpec)
 	UpdateBusinessIndicatorSpec(_ types.NamespacedName, _ []apiworkload.ServiceBusinessIndicatorSpec)
 	UpdateSystemIndicatorSpec(_ types.NamespacedName, _ []apiworkload.ServiceSystemIndicatorSpec)
-	UpdateBusinessIndicatorStatus(_ types.NamespacedName, _ []apiworkload.ServiceBusinessIndicatorStatus)
+	UpdateBusinessIndicatorStatus(_ types.NamespacedName, _ []apiworkload.ServiceBusinessIndicatorStatus, _ []apiworkload.AggPodMetrics)
```
luomingmeng, line 43: the logic of update AggMetrics can use an independent function`UpdateAggMetrics(_ types.NamespacedName, _ []apiworkload.AggPodMetrics)` like [UpdateAggMetrics](https://github.com/tibberzh/katalyst-core/blob/496361573af6d697754b9ffa0b555206bdcc420c/pkg/controller/spd/indicator-plugin/manager.go#L43) in #589

pkg/controller/spd/indicator-plugin/manager.go
```diff
@@ -192,6 +196,8 @@ func (u *IndicatorManager) GetIndicatorStatus(nn types.NamespacedName) *apiworkl
 	return status
 }
 
+func (u *IndicatorManager) UpdateAggMetrics(_ *unstructured.Unstructured) {}
```
luomingmeng, line 199: this function is unused?

pkg/controller/spd/indicator-plugin/manager.go
ddjjia: I didn't notice there's another open mr working on the same thing, so was thinking just leverage the existing status function. Just curious if there're other consideration to make it separate?

luomingmeng: it's beter to separate the function of UpdateAggMetrics and UpdateBusinessIndicatorStatus, because all the update function like `UpdateExtendedIndicatorSpec`, `UpdateBusinessIndicatorSpec` update different aspects of SPD

PR #563, opened 2024-04-29
feat(qrm): use async worker to set extra cg mem limit to avoid thread stucking

csfldf:
feat(qrm): use async worker to set extra cg mem limit to avoid thread stucking

### DIFF
--- a/pkg/util/cgroup/manager/cgroup.go
+++ b/pkg/util/cgroup/manager/cgroup.go
@@ -259,12 +259,12 @@ func GetCPUSetForContainer(podUID, containerId string) (*common.CPUSetStats, err
 }
 
 func DropCacheWithTimeoutForContainer(ctx context.Context, podUID, containerId string, timeoutSecs int, nbytes int64) error {
-	cpusetAbsCGPath, err := common.GetContainerAbsCgroupPath(common.CgroupSubsysMemory, podUID, containerId)
+	memoryAbsCGPath, err := common.GetContainerAbsCgroupPath(common.CgroupSubsysMemory, podUID, containerId)
 	if err != nil {
 		return fmt.Errorf("GetContainerAbsCgroupPath failed with error: %v", err)
 	}
 
-	err = DropCacheWithTimeoutWithRelativePath(timeoutSecs, cpusetAbsCGPath, nbytes)
+	err = DropCacheWithTimeoutAndAbsCGPath(timeoutSecs, memoryAbsCGPath, nbytes)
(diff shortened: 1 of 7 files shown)


cheney-lin (dismissed):

cheney-lin (approved):

pkg/util/cgroup/manager/cgroup.go
```diff
@@ -292,7 +292,54 @@ func DropCacheWithTimeoutWithRelativePath(timeoutSecs int, absCgroupPath string,
...
+	}
+
+	return nil
+}
+
+func SetExtraCGMemLimitWithTimeoutAndRelCGPath(ctx context.Context, relCgroupPath string, timeoutSecs int, nbytes int64) error {
+	memoryAbsCGPath := common.GetAbsCgroupPath(common.CgroupSubsysMemory, relCgroupPath)
+
+	err := SetExtraCGMemLimitWithTimeoutAndAbsCGPath(timeoutSecs, memoryAbsCGPath, nbytes)
+	_ = asyncworker.EmitAsyncedMetrics(ctx, metrics.ConvertMapToTags(map[string]string{
```
zzzzhhb, line 310: should this metric logic be done in asyncworker's scope? what if this method wasn't called by asyncworker?

pkg/util/cgroup/manager/cgroup.go
```diff
@@ -292,7 +292,54 @@ func DropCacheWithTimeoutWithRelativePath(timeoutSecs int, absCgroupPath string,
...
+		//cgv2
+		interfacePath = filepath.Join(absCgroupPath, "memory.max")
+	} else {
+		//cgv1
+		interfacePath = filepath.Join(absCgroupPath, "memory.limit_in_bytes")
+	}
+
+	cmd := fmt.Sprintf("timeout %d echo %d > %s", timeoutSecs, nbytes, interfacePath)
+
+	_, err := exec.Command("bash", "-c", cmd).Output()
```
zzzzhhb, line 339: just curious, why don't we just use io library instead of a bash command?

csfldf: because we want to reuse TIMEOUT(1)  user commands to trigger timeout to prevent stucking

zzzzhhb: get

pkg/util/cgroup/manager/cgroup.go
csfldf: [TODO]
1. use async worker framework to record succees information for works
2. pass and use normal emitter to record specic information for works

zzzzhhb (approved):

PR #572, opened 2024-05-06
feat(general): audit manager framework

zzzzhhb:
This PR adds a framework to gather all the control actions(like applying cgroups, system calls and etc)executed by katalyst then send them to a data warehouse. This mechanism aims to provide audit data for users. If there is a need to integrate a 3rd-party storage, we can simply implement a `Sink` interface to send data to it.

### DIFF
--- a/pkg/util/eventbus/event_bus.go
+++ b/pkg/util/eventbus/event_bus.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2022 The Katalyst Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package eventbus
+
+import (
+	"fmt"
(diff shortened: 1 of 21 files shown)


pkg/util/audit/manager.go
```diff
@@ -0,0 +1,67 @@
+/*
```
waynepeking348, line 1: maybe it is better to put this manager in `pkg/agent` rather than `pkg/util`?

waynepeking348 (dismissed):

waynepeking348 (dismissed):

luomingmeng (approved):

waynepeking348 (approved):

luomingmeng (approved):

pkg/util/eventbus/event_bus.go
```diff
@@ -0,0 +1,164 @@
...
+			eventHandlers: make(map[string]*eventHandler),
+		}
+		go e.topics[topic].Run()
+		general.Infof("register new topic: %v", topic)
+		return e.topics[topic]
+	}
+}
+
+func (e *eventBus) Publish(topic string, event interface{}) error {
+	e.mutex.RLock()
```
csfldf, line 144: is it better to make mutex as a topic level lock to aviod needless contention in this lock?

zzzzhhb: @csfldf  Maybe not, this lock aims to prevent races on the `topics` map, since publishing only need to read the map so we use a `RLock` here. The contention will only occurs when a new topic is added.

csfldf (approved):

waynepeking348 (approved):

PR #584, opened 2024-05-16
make malachite server port configurable

cccaoke:
#### What type of PR is this?
Enhancements

malachite metrics provisoner using default port(9002) that may be used by others, by overriding the Rocket config file, but not katalyst.

make malachite server port configurable


cccaoke:
the conflict has been resolved，pls take a look @waynepeking348 @nightmeng

PR #585, opened 2024-05-16
fix: ignore not existed cgroup paths

cheney-lin:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/cmd/katalyst-agent/app/options/global/base.go
+++ b/cmd/katalyst-agent/app/options/global/base.go
@@ -52,6 +52,7 @@ type BaseOptions struct {
 
 	ReclaimRelativeRootCgroupPath string
 	GeneralRelativeCgroupPaths    []string
+	OptionalRelativeCgroupPaths   []string
 
 	// configurations for kubelet
 	KubeletReadOnlyPort      int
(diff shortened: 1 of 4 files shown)


nightmeng (approved):

pkg/metaserver/agent/metric/provisioner/malachite/provisioner.go
```diff
@@ -169,6 +169,11 @@ func (m *MalachiteMetricsProvisioner) updateCgroupData() error {
 	cgroupPaths := []string{m.baseConf.ReclaimRelativeRootCgroupPath, common.CgroupFsRootPathBurstable, common.CgroupFsRootPathBestEffort}
 	errList := make([]error, 0)
 	for _, path := range m.baseConf.GeneralRelativeCgroupPaths {
```
nightmeng, line 171: The OptionalRelativeCgroupPaths is a general configuration of malichite, and other components may depend on this configuration in the future. It is not reasonable to ignore non-existent cgroup paths by default. Please add an OptionalRelativeCgroupPaths for this case.

nightmeng (approved):

zzzzhhb (approved):

cmd/katalyst-agent/app/options/global/base.go
```diff
@@ -52,6 +52,7 @@ type BaseOptions struct {
 
 	ReclaimRelativeRootCgroupPath string
 	GeneralRelativeCgroupPaths    []string
+	OptionalRelativeCgroupPaths   []string
```
waynepeking348, line 55: why options are defined in `BaseOptions` but configs are defined in `MalachiteConfiguration`?

waynepeking348 (approved):

PR #592, opened 2024-05-23
feat: add MaxNodeUtilizationPercent in ReclaimedResourceConfiguration

WangZzzhe:
#### What type of PR is this?
Enhancements

Add parameter MaxNodeUtilizationPercent，which will be converted to reservedForAllocated to limit the total amount of reclaimed resources.

PR should be updated after https://github.com/kubewharf/katalyst-api/pull/85 is merged

### DIFF
--- a/pkg/config/agent/dynamic/adminqos/reclaimedresource/reclaimedresource_base.go
+++ b/pkg/config/agent/dynamic/adminqos/reclaimedresource/reclaimedresource_base.go
@@ -18,10 +18,12 @@ package reclaimedresource
 
 import (
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 
 	"github.com/kubewharf/katalyst-core/pkg/config/agent/dynamic/adminqos/reclaimedresource/cpuheadroom"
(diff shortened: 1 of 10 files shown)


pkg/config/agent/dynamic/adminqos/reclaimedresource/reclaimedresource_base.go
```diff
@@ -72,8 +75,50 @@ func (c *ReclaimedResourceConfiguration) ApplyConfiguration(conf *crd.DynamicCon
...
+				c.MaxNodeUtilizationPercent[resourceName] = value
+			}
+		}
 	}
 
 	c.CPUHeadroomConfiguration.ApplyConfiguration(conf)
 	c.MemoryHeadroomConfiguration.ApplyConfiguration(conf)
 }
+
+func (c *ReclaimedResourceConfiguration) GetReservedResourceForAllocate(nodeResourceList v1.ResourceList) v1.ResourceList {
```
luomingmeng, line 90: It may be better for GetReservedResourceForAllocate to use resource name and quantity as two separate input parameters.

PR #601, opened 2024-06-01
feat(pap) : power aware plugin

h-w-chen:
#### What type of PR is this?
Feature

It introduces the new feature - power aware plugin.

Its general architecture:
* main body in sysadvisor;
* CPU frequency RAPL capping utilitity lib wrapper in external manager;
* workload evict interaction is, as an interim means, abstracted as interface of podKiller.PodKiller;

By default this plugin is enabled; in order to disable it, 
```bash
--power-aware-disabled=true
```
This plugin supports DryRun mode; to enable DryRun
```bash
--power-aware-dryrun=true
```
NA - new feature

Pod Evict API is not in place yet; for interim purpose, this PR assumes podKiller.PodKiller interface (not finalized and could change), provides a dummy no-op implementation, which is for now to facilitate testing and in the future to be replaced with a real meaningful implementation.


pkg/agent/sysadvisor/plugin/poweraware/component/evictor.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (l loadEvictor) isBE(pod *v1.Pod) bool {
+	qosLevel, err := l.qosConfig.GetQoSLevelForPod(pod)
+	if err != nil {
+		// unknown, not BE anyway
+		return false
+	}
+
+	return qosLevel == apiconsts.PodAnnotationQoSLevelSharedCores ||
```
h-w-chen, line 60: share-core type is NOT BE, is it?

pkg/agent/sysadvisor/plugin/poweraware/component/reconsiler.go
zzzzhhb, line 1: reconsiler?typo?

pkg/agent/sysadvisor/plugin/poweraware/component/evictor.go
zzzzhhb: no, shared-cores is GT. The `apiconsts.PodAnnotationQoSLevelReclaimedCores`  is equal to BE.  But in katalyst, there is no concept of BE. We only have the four types of QoS, so please just use the `apiconsts.PodAnnotationQoSLevelReclaimedCores`.

h-w-chen: thanks for verification. will change code to reflect that semantic.

pkg/agent/sysadvisor/plugin/poweraware/component/reconsiler.go
h-w-chen: yes typo. will rectify it.

PR #605, opened 2024-06-03
refactor(async): support limited async workers

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/util/asyncworker/async_limited_workers.go
+++ b/pkg/util/asyncworker/async_limited_workers.go
@@ -1,0 +1,286 @@
+/*
+Copyright 2022 The Katalyst Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package asyncworker
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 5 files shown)


pkg/util/asyncworker/async_limited_workers.go
```diff
@@ -0,0 +1,245 @@
...
+		"params", work.Params,
+		"deliveredAt", work.DeliveredAt)
+
+	status, ok := alw.activeQ[work.Name]
+	if ok {
+		if policy == DuplicateWorkPolicyDiscard {
+			general.Infof("work %v already exists, discard new work", work.Name)
+			return nil
+		}
+		alw.backoffQ[work.Name] = work
```
csfldf, line 80: is it better to add log for the key overrides in backoffQ?

pkg/util/asyncworker/async_limited_workers.go
```diff
@@ -0,0 +1,245 @@
...
+		alw.waitQ.Add(work)
+		delete(alw.backoffQ, completedWork.Name)
+	}
+	delete(alw.activeQ, completedWork.Name)
+
+	alw.cond.Broadcast()
+}
+
+func (alw *AsyncLimitedWorkers) Start(stopCh <-chan struct{}) error {
+	go wait.Until(alw.handle, 10*time.Second, stopCh)
```
csfldf, line 243: should we 
1. add another goroutine to select if stopCh closed, if so, triggering alw.cond.Broadcast()
2. perceive stopCh closed in doPoll -> poll -> handle, then stopping loop in handle

pkg/util/asyncworker/async_limited_workers.go
```diff
@@ -0,0 +1,245 @@
...
+		return "", fmt.Errorf("invalid work type")
+	}
+	return w.Name, nil
+}
+
+func workLessFunc(w1, w2 interface{}) bool {
+	return w1.(*Work).DeliveredAt.Before(w2.(*Work).DeliveredAt)
+}
+
+func NewAsyncLimitedWorkers(name string, limit int, emitter metrics.MetricEmitter) *AsyncLimitedWorkers {
```
csfldf, line 45: it's better to add a metric to monitor the duration that a work pending in queue (both backoff and wait), if it's complicated we can add a TODO comment and implement it in the future.

csfldf (approved):

PR #629, opened 2024-06-17
feat(cpu): add cpu plugin for vci

gaohuatao-1:
#### What type of PR is this?
Features

To support  customized cpu policy for VCI with pod level resources.

### DIFF
--- a/pkg/agent/qrm-plugins/cpu/vcipolicy/policy.go
+++ b/pkg/agent/qrm-plugins/cpu/vcipolicy/policy.go
@@ -1,0 +1,391 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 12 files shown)


pkg/agent/qrm-plugins/cpu/vcipolicy/policy.go
```diff
@@ -0,0 +1,392 @@
...
+	pluginapi "k8s.io/kubelet/pkg/apis/resourceplugin/v1alpha1"
+)
+
+const (
+	// cpuPluginStateFileName is the name of cpu plugin state file.
+	cpuPluginStateFileName = "cpu_plugin_state"
+)
+
+// NativePolicy is a policy compatible with Kubernetes native semantics and is used in topology-aware scheduling scenarios.
+type VCIStaticPolicy struct {
```
caohe, line 51: Is there a more generic name than "VCI" policy?

WangZzzhe: Maybe some constants containing 'vci.volcengine.com' can be passed through args to avoid the appearance of these non-public naming in the code ?

caohe: Is this policy suitable for open source?

PR #645, opened 2024-07-05
fix(advisor): optimize lw health check

nightmeng:
#### What type of PR is this?
Enhancements

Support short time health check

### DIFF
--- a/pkg/util/general/healthz.go
+++ b/pkg/util/general/healthz.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"sync"
 	"time"
+
+	"k8s.io/klog/v2"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/util/general/healthz.go
```diff
@@ -122,9 +154,35 @@ func RegisterReportCheck(name string, autoRecoverPeriod time.Duration) {
...
 	}
 }
 
+func RegisterShortTimeReportCheck(name string, autoRecoverPeriod time.Duration) {
+	healthzCheckLock.Lock()
+	defer healthzCheckLock.Unlock()
+
+	count := 1
+	current, ok := healthzCheckMap[HealthzCheckName(name)]
+	if ok {
```
luomingmeng, line 167: if healthz check already exists, just update the `autoRecoverPeriod` and return may be beter

luomingmeng (approved):

xu282934741 (approved):

PR #666, opened 2024-08-01
support clear residual net class

luomingmeng:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/qrm-plugins/network/staticpolicy/policy.go
+++ b/pkg/agent/qrm-plugins/network/staticpolicy/policy.go
@@ -57,6 +57,11 @@ const (
 
 	// IPsSeparator is used to split merged IPs string
 	IPsSeparator = ","
+
+	// residualNetClassTTL is the ttl of residual net class in cgroupv2
+	residualNetClassTTL = 30 * time.Minute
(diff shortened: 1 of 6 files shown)


pkg/agent/qrm-plugins/network/staticpolicy/policy.go
```diff
@@ -769,6 +782,10 @@ func (p *StaticPolicy) applyNetClass() {
 			}(string(pod.UID), container.Name, netClsData)
 		}
 	}
+
+	if p.CgroupV2Env {
```
caohe, line 786: Do we need to clear residual net class data for cgroup v1?

caohe (approved):

pkg/agent/qrm-plugins/network/staticpolicy/policy.go
luomingmeng: In cgroup v1, it is not necessary to clear residual net class because the net class is set within the cgroup path. This will be automatically cleared when the pod is terminated by kubelet.

xu282934741 (approved):

csfldf (approved):

PR #674, opened 2024-08-14
fix(reporter): fix race condition regarding orm and reporter start order

caohe:
#### What type of PR is this?

Bug fixes

Fix race condition regarding orm and reporter start order.

In ORM mode, kubelet reporter plugin get topology info from ORM's out-of-band PodResources API, rather than kubelet's native PodResources API. If reporter starts before ORM, it can not get the socket of PodResources Server.

Error log:
```
E0807 17:58:09.245252       1 manager.go:403] GetReportContentResponse from kubelet-reporter-plugin Endpoint failed with error: get numa topology status from adapter failed: list pod from pod resource server failed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix: missing address"
```

So I extracted the PodResources client to a common util and added a retry mechanism.

### DIFF
--- a/pkg/util/kubelet/podresources/client.go
+++ b/pkg/util/kubelet/podresources/client.go
@@ -18,14 +18,106 @@ package podresources
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"k8s.io/klog/v2"
 	v1 "k8s.io/kubelet/pkg/apis/podresources/v1"
 	"k8s.io/kubernetes/pkg/kubelet/util"
+
+	"github.com/kubewharf/katalyst-core/pkg/util/general"
 )
+
+const (
+	podResourcesClientTimeout              = 10 * time.Second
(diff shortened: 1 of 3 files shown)


luomingmeng (approved):

pkg/util/kubelet/podresources/client.go
```diff
@@ -18,15 +18,107 @@ package podresources
...
+)
+
+const (
+	podResourcesClientTimeout              = 10 * time.Second
+	podResourcesClientMaxMsgSize           = 1024 * 1024 * 16
+	checkPodResourcesSocketInterval        = 100 * time.Millisecond
+	checkPodResourcesSocketTimeoutDuration = 10 * time.Second
 )
 
+var errNotImplemented = errors.New("pod resources client not initialized")
```
luomingmeng, line 42: rename errNotImplemented to errNotInitialized

caohe: done

luomingmeng (approved):

WangZzzhe (approved):

csfldf (approved):

PR #694, opened 2024-09-05
feat:  report NUMA cpu usage

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/util/metric/store.go
+++ b/pkg/util/metric/store.go
@@ -39,16 +39,16 @@ type MetricData struct {
 type MetricStore struct {
 	mutex sync.RWMutex
 
-	nodeMetricMap             map[string]MetricData                                  // map[metricName]data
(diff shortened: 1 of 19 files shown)


pkg/util/metric/store.go
```diff
@@ -248,6 +248,22 @@ func (c *MetricStore) GetContainerNumaMetric(podUID, containerName, numaNode, me
 	return MetricData{}, errors.New(fmt.Sprintf("[MetricStore] empty map, metric=%v, podUID=%v, containerName=%v, numaNode=%v", metricName, podUID, containerName, numaNode))
 }
 
+func (c *MetricStore) GetContainerNumaMetrics(podUID, containerName, metricName string) (map[string]MetricData, error) {
+	c.mutex.RLock()
```
luomingmeng, line 252: why not translate `map[string]MetricData` to `map[int]MetricData` here directly

luomingmeng (approved):

PR #697, opened 2024-09-11
refactor(qrm): refactor the state module for enhanced extensibility

xu282934741:

if we want to add a new strategy, this refactoring will be helpful.

### DIFF
--- a/pkg/agent/qrm-plugins/cpu/dynamicpolicy/state/util.go
+++ b/pkg/agent/qrm-plugins/cpu/dynamicpolicy/state/util.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/klog/v2"
 
 	apiconsts "github.com/kubewharf/katalyst-api/pkg/consts"
-	"github.com/kubewharf/katalyst-core/pkg/agent/qrm-plugins/cpu/consts"
 	cpuconsts "github.com/kubewharf/katalyst-core/pkg/agent/qrm-plugins/cpu/consts"
 	"github.com/kubewharf/katalyst-core/pkg/util/general"
 	"github.com/kubewharf/katalyst-core/pkg/util/machine"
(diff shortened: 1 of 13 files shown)


pkg/agent/qrm-plugins/cpu/nativepolicy/util/util.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/util/machine"
 )
 
+// GenerateMachineStateFromPodEntries for native policy
 func GenerateMachineStateFromPodEntries(topology *machine.CPUTopology, podEntries state.PodEntries) (state.NUMANodeMap, error) {
```
luomingmeng, line 26: Begin with a lowercase letter like `generateMachineStateFromPodEntries`?

xu282934741: @luomingmeng NewNativePolicy function in package nativepolicy also need to call GenerateMachineStateFromPodEntries function

pkg/agent/qrm-plugins/cpu/dynamicpolicy/state/util.go
```diff
@@ -490,3 +430,67 @@ func checkCPUSetMap(map1, map2 map[int]machine.CPUSet) bool {
 func CPUPreciseCeil(request float64) int {
 	return int(math.Ceil(float64(int(request*1000)) / 1000))
 }
+
+// GenerateMachineStateFromPodEntriesByPolicy returns NUMANodeMap for given resource based on
+// machine info and reserved resources along with existed pod entries and policy name
+func GenerateMachineStateFromPodEntriesByPolicy(topology *machine.CPUTopology, podEntries PodEntries, policyName string) (NUMANodeMap, error) {
```
luomingmeng, line 436: Why the GenerateMachineStateFromPodEntriesByPolicy is in dynamicpolicy state util, and not a  common place for both dynamicpolicy and nativepolicy

xu282934741: @luomingmeng Good Question. Since the GenerateMachineStateFromPodEntries method depends on too many other methods and structures within the state package, extracting it as a standalone utility would require extracting the entire state package. However, doing so would incur a significant refactoring cos

luomingmeng: ok, add some todo comments for future refinement

xu282934741: @luomingmeng done

PR #720, opened 2024-11-12
feat (sysadvisor): get power data served by metric store sourced from malachite realtime metric server

h-w-chen:
#### What type of PR is this?
Feature: power aware advisor uses realtime power data originally from malachite. It has 3 major parts:
* malachite client method to get power stats from metric service path realtime/power;
* malachite_realtime provisioner which samples (in higher frequency) using malachite client and saves power metric in metric store;
* a power reader adapted to get power data out of metric store; 

part of power node agent, it is needed for power aware plugin to get the current power usage and hence identify whether further measures required to limit the power consumption comparing with the desired power target.

NA

### DIFF
--- a/pkg/metaserver/agent/metric/provisioner/malachite/realtime_provisioner.go
+++ b/pkg/metaserver/agent/metric/provisioner/malachite/realtime_provisioner.go
@@ -80,7 +80,7 @@ func (m *MalachiteRealtimeMetricsProvisioner) Run(ctx context.Context) {
 func (m *MalachiteRealtimeMetricsProvisioner) checkMalachiteHealthy() bool {
 	_, err := m.malachiteClient.GetPowerData()
 	if err != nil {
-		klog.Errorf("[malachite] malachite realtime is unhealthy: %v", err)
+		klog.Errorf("[malachite_realtime] malachite realtime is unhealthy: %v", err)
 		_ = m.emitter.StoreInt64(metricsNamMalachiteRealtimeUnHealthy, 1, metrics.MetricTypeNameRaw)
 		return false
 	}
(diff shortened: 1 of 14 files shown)


pkg/metaserver/agent/metric/provisioner/malachite/realtime_provisioner.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const (
+	metricsNamMalachiteRealtimeUnHealthy        = "malachite_realtime_unhealthy"
+	malachiteRealtimeProvisionerHealthCheckName = "malachite_realtime_provisioner_sample"
+	malachiteRealtimeProvisionTolerationTime    = 5 * time.Second
+	metricsNameMalachiteGetRealtimePowerFailed  = "malachite_get_realtime_power_failed"
+)
+
+func NewMalachiteRealtimeMetricsProvisioner(baseConf *global.BaseConfiguration, metricConf *metaserver.MetricConfiguration,
```
gary-lgy, line 45: Do we plan to pass the sampling interval from command line arguments?

h-w-chen: yes the arg should be like "--metric-provisioner-intervals=malachite_reltime=1" (this is to specify the interval to be 1 second instead of the default value)

h-w-chen: change code to set the default interval for malachite_realtime to 1 sec. so no need for explicit arg unless the 1 sec is not desired.

pkg/metaserver/agent/metric/provisioner/malachite/realtime_provisioner.go
```diff
@@ -0,0 +1,131 @@
...
+		klog.Errorf("[malachite] malachite realtime is unhealthy: %v", err)
+		_ = m.emitter.StoreInt64(metricsNamMalachiteRealtimeUnHealthy, 1, metrics.MetricTypeNameRaw)
+		return false
+	}
+
+	return true
+}
+
+func (m *MalachiteRealtimeMetricsProvisioner) sample(ctx context.Context) {
+	klog.V(4).Infof("[malachite] heartbeat")
```
xu282934741, line 92: Would it be more appropriate to print the log here as "[malachite_realtime] heartbeat" to avoid a conflict with another malachite provisioner

h-w-chen: yes will update log as suggested

xu282934741 (approved):

PR #721, opened 2024-11-14
fix(*): empty cpuset.mems will lead to failure of tmo

cheney-lin:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
+++ b/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
@@ -736,13 +736,9 @@ func (p *DynamicPolicy) handleAdvisorMemoryOffloading(_ *config.Configuration,
 		}
 	}
 
-	cpuSetStats, err := cgroupmgr.GetCPUSetWithAbsolutePath(absCGPath)
+	_, mems, err := cgroupmgr.GetEffectiveCPUSetWithAbsolutePath(absCGPath)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
```diff
@@ -736,11 +736,7 @@ func (p *DynamicPolicy) handleAdvisorMemoryOffloading(_ *config.Configuration,
...
 	}
 
-	cpuSetStats, err := cgroupmgr.GetCPUSetWithAbsolutePath(absCGPath)
-	if err != nil {
-		return fmt.Errorf("GetCPUSetWithAbsolutePath failed with error: %v", err)
-	}
-	mems, err := machine.Parse(cpuSetStats.Mems)
+	_, mems, err := cgroupmgr.GetEffectiveCPUSetWithAbsolutePath(absCGPath)
 	if err != nil {
 		return fmt.Errorf("parse cpuSetStats failed with error: %v", err)
```
xu282934741, line 741: @cheney-lin Corresponding changes to the log content.

pkg/util/cgroup/manager/cgroup.go
```diff
@@ -514,3 +518,49 @@ func MemoryOffloadingWithAbsolutePath(ctx context.Context, absCgroupPath string,
 
 	return err
 }
+
+func IsCgroupPath(path string) bool {
+	var fstat syscall.Statfs_t
+	err := syscall.Statfs(path, &fstat)
+	if err != nil {
+		return false
```
xu282934741, line 526: Should we print the error here before returning false?

PR #728, opened 2024-11-25
fix (sysadvisor): pap - fixes of defects identified in integration validation

h-w-chen:
#### What type of PR is this?
Bug fixes

this PR has 3 fixes:
1.  explicit internal-op over autonomy in last 2 minutes (behavior change);
2. power cap sock file folders creation in case it is not there;
3. split power cap server and plugin to avoid recursive reference chain via plugin wrapper

### DIFF
--- a/pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
+++ b/pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
@@ -57,7 +57,10 @@ func (p ruleBasedPowerStrategy) RecommendAction(actualWatt int,
 
 	if ttl <= time.Minute*2 {
 		// whatever valid alert, power capping should do in short of 2 minutes
-		return action.PowerAction{Op: spec.InternalOpFreqCap, Arg: desiredWatt}
+		// this the default behavior unless instructed explicitly by internal op
+		if spec.InternalOpAuto == internalOp {
+			return action.PowerAction{Op: spec.InternalOpFreqCap, Arg: desiredWatt}
+		}
 	}
 
 	op := internalOp
--- a/pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
+++ b/pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
@@ -55,14 +55,6 @@ func (p ruleBasedPowerStrategy) RecommendAction(actualWatt int,
 		return action.PowerAction{Op: spec.InternalOpNoop, Arg: 0}
 	}
 
-	if ttl <= time.Minute*2 {
-		// whatever valid alert, power capping should do in short of 2 minutes
-		// this the default behavior unless instructed explicitly by internal op
-		if spec.InternalOpAuto == internalOp {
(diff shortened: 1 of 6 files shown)


pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
```diff
@@ -57,7 +57,10 @@ func (p ruleBasedPowerStrategy) RecommendAction(actualWatt int,
 
 	if ttl <= time.Minute*2 {
 		// whatever valid alert, power capping should do in short of 2 minutes
-		return action.PowerAction{Op: spec.InternalOpFreqCap, Arg: desiredWatt}
+		// this the default behavior unless instructed explicitly by internal op
```
gary-lgy, line 60: Should this be part of the `autoAction` function?

pkg/agent/sysadvisor/plugin/poweraware/evictor/server/evictor_service.go
```diff
@@ -110,16 +110,9 @@ func (p *powerPressureEvictServer) Start() error {
...
-
-	if !p.started {
-		general.InfofV(6, "pap: power pressure eviction server already stopped")
-		return nil
-	}
-
-	p.started = false
-	return p.service.Stop()
+	// since there is no resource other than the underlying PluginRegistrationWrapper-guarded grpc service,
+	// which will be cleanup by PluginRegistrationWrapper itself, it has nothing to clean up here
```
gary-lgy, line 114: `PluginRegistrationWrapper` has an internal stop channel that controls the stopping of the server. How do we close the stop channel without calling `Stop` on the wrapper?

pkg/agent/sysadvisor/plugin/poweraware/advisor/action/strategy/strategy.go
h-w-chen: yes it is actually. will update code NOT to return directly here, and let autoAction to fill up the needed detail.

pkg/agent/sysadvisor/plugin/poweraware/evictor/server/evictor_service.go
h-w-chen: good point! the current implementation has a recursive reference, which leads to recursive calls of Stop. Will reconstruct code to break the loop.

pkg/agent/sysadvisor/plugin/poweraware/evictor/server/evictor_service.go
```diff
@@ -43,24 +42,47 @@ const (
 var errPowerPressureEvictionPluginUnavailable = errors.New("power pressure eviction plugin is unavailable")
 
 type powerPressureEvictServer struct {
+	plugin  *powerPressureEvictPlugin
```
gary-lgy, line 45: Where is `plugin` initialized?

h-w-chen: thanks for catching this - will add it as server object field

gary-lgy (approved):

PR #734, opened 2024-12-04
support cpu codename aware indicator

dongmx:


pkg/consts/metric.go
```diff
@@ -176,6 +176,7 @@ const (
 	MetricCPUUsageRatio    = "cpu.usage.ratio.cpu"
 	MetricCPUSysUsageRatio = "cpu.sys.usage.ratio.cpu"
 	MetricCPUIOWaitRatio   = "cpu.iowait.ratio.cpu"
+	MetricCPUCodeName      = "cpu.code.name"
```
xu282934741, line 179: @dongmx For metrics that are not particularly common, it is recommended to add some comments to describe the meaning of the metric.

xu282934741 (approved):

PR #742, opened 2024-12-12
refactor cnr taint

luomingmeng:
#### What type of PR is this?

Enhancements


pkg/controller/lifecycle/agent-healthz/helper/taint_cnr.go
```diff
@@ -43,18 +44,21 @@ const (
 
 const TaintNameNoScheduler = "TaintNameNoScheduler"
 
-var TaintNoScheduler = &apis.Taint{
-	Key:    corev1.TaintNodeUnschedulable,
-	Effect: apis.TaintEffectNoScheduleForReclaimedTasks,
+var TaintNoScheduler = apis.Taint{
```
xu282934741, line 47: Similar to the above，TaintReclaimedCoresNoSchedule more appropriate here?

pkg/controller/lifecycle/agent-healthz/helper/taint_cnr.go
```diff
@@ -43,18 +44,21 @@ const (
 
 const TaintNameNoScheduler = "TaintNameNoScheduler"
```
xu282934741, line 45: Would it be more appropriate to call it TaintNameReclaimedCoresNoSchedule here？

xu282934741 (approved):

PR #759, opened 2025-01-11
WIP: metrics(sysadvisor) - power aware advisor emits metrics of essential ops

h-w-chen:


pkg/agent/sysadvisor/plugin/poweraware/metric/error_code.go
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metric
+
+const (
+	ErrorCodeInitFailure                  = 9505001
```
gary-lgy, line 20: Why don't we emit these as strings that describe the error?

pkg/agent/sysadvisor/plugin/poweraware/metric/emit_metric.go
```diff
@@ -0,0 +1,85 @@
...
+		})...,
+	)
+}
+
+func EmitPowerCapReset(emitter metrics.MetricEmitter) {
+	_ = emitter.StoreInt64(metricPowerCappingReset, 1, metrics.MetricTypeNameCount)
+}
+
+func EmitErrorCode(emitter metrics.MetricEmitter, errorCode int) {
+	_ = emitter.StoreInt64(metricPowerError, 1, metrics.MetricTypeNameRaw,
```
gary-lgy, line 80: Should we use counter type for this metric? Using raw type means the value will always be 1 or 0 and does not reflect the number of errors.

pkg/agent/sysadvisor/plugin/poweraware/metric/error_code.go
h-w-chen: ok will come up with error string list (which is much human friendlier) replacing the numbers

pkg/agent/sysadvisor/plugin/poweraware/metric/emit_metric.go
h-w-chen: counter type here is the right choice - will change as suggest

PR #791, opened 2025-02-25
feat: add cpu_usage_ratio as target indicator of dedicated_cores region

cheney-lin:
#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/helper/pid.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/helper/pid.go
@@ -75,6 +75,10 @@ func (c *PIDController) Adjust(controlKnob, target, current float64) float64 {
 			kp = c.params.Kpn
 			kpSign = -1
 		}
+		// The default value of kp is evaluated on the machine which has 96 Cores.
+		// In an actual scenario, if kp remains unchanged, when the region is large, the control will be excessively sluggish.
+		// So here we scale kp according to the region size.
+		kp *= c.resourceEssentials.ResourceUpperBound
 
 		if errorRate >= 0 {
 			kdSign = kpSign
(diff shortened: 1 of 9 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/helper/pid.go
```diff
@@ -75,6 +75,7 @@ func (c *PIDController) Adjust(controlKnob, target, current float64) float64 {
 			kp = c.params.Kpn
 			kpSign = -1
 		}
+		kp *= c.resourceEssentials.ResourceUpperBound / 100
```
luomingmeng, line 78: Why is it necessary to combine kp with resourceUpperBound and then divide by 100? Is it compatible with all scenarios?

pkg/consts/metric.go
```diff
@@ -155,6 +155,7 @@ const (
 
 // System numa metrics
 const (
+	MetricCPUNrNuma           = "cpu.nr.numa"
```
luomingmeng, line 158: Does it refer to the number of NUMA nodes, which can be obtained from machineInfo?

cheney-lin: The topology of some machines may not be completely symmetric in terms of CPU and memory. For instance, situations like CXL (Compute Express Link) or fake NUMA (Non-Uniform Memory Access) will have this characteristic.

pkg/agent/sysadvisor/plugin/qosaware/resource/helper/pid.go
cheney-lin: yes，the purpose is to make k proportional to the region.

luomingmeng (approved):

luomingmeng (approved):

PR #807, opened 2025-03-26
refactor(*): move page don't return error if a process has already ex…

luomingmeng:
…ited

#### What type of PR is this?

Enhancements

### DIFF
--- a/pkg/agent/qrm-plugins/memory/dynamicpolicy/util_linux_amd64.go
+++ b/pkg/agent/qrm-plugins/memory/dynamicpolicy/util_linux_amd64.go
@@ -21,11 +21,14 @@ package dynamicpolicy
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/agent/qrm-plugins/memory/dynamicpolicy/util_linux_amd64.go
```diff
@@ -491,8 +499,10 @@ func getNumasFreeMemRatio(systemNodeDir string, numas []int) (map[int]int, error
 func getProcessPageStats(procDir string, pid int) (*smapsInfo, error) {
 	smapFile := filepath.Join(procDir, fmt.Sprintf("%d/smaps", pid))
 	lines, err := general.ReadFileIntoLines(smapFile)
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
xu282934741, line 502: The logic of this code is a bit obscure, can it be changed to the following？
if err != nil {
    if os.IsNotExist(err) {
        return nil, nil
    }
    return nil, fmt.Errorf("failed to ReadLines(%s), err %v", smapFile, err)
}

xu282934741 (approved):

PR #832, opened 2025-04-21
enhance(advisor): power cap service supports GetAdvice

h-w-chen:
#### What type of PR is this?
Enhancements - power cap advisor supports GetAdvice unary GRPC protocol, with long polling ability

This PR implements GetAdvice method of new advisor-qrm communication protocol. To avoid communication overhead, it  leverages long-polling so as to reduce the client requests resulting empty responses.

### DIFF
--- a/pkg/agent/sysadvisor/plugin/poweraware/capper/server/service.go
+++ b/pkg/agent/sysadvisor/plugin/poweraware/capper/server/service.go
@@ -42,14 +42,14 @@ const (
 	// ServiceNamePowerCap also is the unix socket name of the server is listening on
 	ServiceNamePowerCap = "node_power_cap"
 
-	metricServerGetAdviceCalled = "pap_get_advice_called"
-
 	pollingTimeout = time.Second * 60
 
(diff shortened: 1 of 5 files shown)


pkg/agent/sysadvisor/plugin/poweraware/capper/server/service.go
```diff
@@ -39,15 +41,51 @@ import (
...
 )
 
+type longPoller struct {
+	timeout time.Duration
+	readyCh chan struct{}
+}
+
+func (l *longPoller) setReady() {
+	// chan close unblocks all waiting for the ready signal
+	close(l.readyCh)
```
JulyWindK, line 62: Is it necessary to add a judgment here to prevent closing a closed chan.

h-w-chen: in our case it is not needed, as the only place to close is this method, and followed by the chan re-assignment, plus  no multiple method calls are possible (power cap advisor as the only caller).

JulyWindK (approved):

PR #840, opened 2025-04-29
fix(tmo): remove target size zero check to fix swap not off

yadzhang:
#### What type of PR is this?
Bug fixes

remove memory advisor the check for offloading target size less than or equal to zero to fix swap not off when target pod tmo is disabled.

### DIFF
--- a/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
+++ b/pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
@@ -863,11 +863,6 @@ func (p *DynamicPolicy) handleAdvisorMemoryOffloading(_ *config.Configuration,
 ) error {
 	var absCGPath string
 	var memoryOffloadingWorkName string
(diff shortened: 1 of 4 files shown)


pkg/agent/qrm-plugins/memory/dynamicpolicy/policy_advisor_handler.go
```diff
@@ -898,6 +893,16 @@ func (p *DynamicPolicy) handleAdvisorMemoryOffloading(_ *config.Configuration,
 		}
 	}
 
+	memoryOffloadingSizeInBytes := calculationInfo.CalculationResult.Values[string(memoryadvisor.ControlKnowKeyMemoryOffloading)]
+	memoryOffloadingSizeInBytesInt64, err := strconv.ParseInt(memoryOffloadingSizeInBytes, 10, 64)
+	if err != nil {
+		return fmt.Errorf("parse %s: %s failed with error: %v", memoryadvisor.ControlKnowKeyMemoryOffloading, memoryOffloadingSizeInBytes, err)
+	}
+
+	if memoryOffloadingSizeInBytesInt64 <= 0 {
```
luomingmeng, line 902: Add logs to indicate when memory offloading is skipped.

luomingmeng (approved):

PR #846, opened 2025-05-09
chore(*): fix kubelet endpoint do not support ipv6

cccaoke:
#### What type of PR is this?
Bug fixes

for ipv6 only enviroment, katalyst does not work

ipv6 compatible when requesting Kubelet


cccaoke:
cc @luomingmeng @caohe

luomingmeng (approved):

PR #849, opened 2025-05-12
feat(npd): add npd into cnc

Bourbon-Wang:


pkg/controller/kcc/kcct.go
```diff
@@ -338,6 +346,31 @@ func (k *KatalystCustomConfigTargetController) syncKCCTs(gvr metav1.GroupVersion
 		return utilerrors.NewAggregate(errors)
```
luomingmeng, line 346: There might be a bug that should return an error if the length of `errors` is greater than 0, and return nil if the length of `targetResources` equals 0.

pkg/controller/kcc/target/target_handler.go
```diff
@@ -141,6 +149,17 @@ func (k *KatalystCustomConfigTargetHandler) RangeGVRTargetAccessor(f func(gvr me
 	}
 }
 
+func (k *KatalystCustomConfigTargetHandler) registerDefaultGVR(gvrStr string) error {
+	gvr, _ := schema.ParseResourceArg(gvrStr)
+	if gvr == nil {
+		return fmt.Errorf("parse to GroupVersionResource failed")
+	}
+
+	k.kccConfig.ValidAPIGroupSet.Insert(gvr.Group)
```
luomingmeng, line 158: Avoid directly inserting ValidAPIGroupSet. Instead, consider introducing a local member within `KatalystCustomConfigTargetHandler` that mirrors the ValidAPIGroupSet configuration.

luomingmeng (approved):

PR #877, opened 2025-06-20
feat(gpu): add static policy implementation for GPU resource management

luomingmeng:
#### What type of PR is this?

Features
This commit introduces a new static policy implementation for GPU resource management, including:
- GPU topology provider and state management
- Static policy implementation with allocation and deallocation logic
- Integration with existing QRM framework
- Metrics and health checks for GPU resource management

### DIFF
--- a/pkg/agent/qrm-plugins/gpu/staticpolicy/policy.go
+++ b/pkg/agent/qrm-plugins/gpu/staticpolicy/policy.go
@@ -19,6 +19,7 @@ package staticpolicy
 import (
 	"context"
 	"fmt"
+	"math"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/agent/qrm-plugins/gpu/staticpolicy/policy.go
```diff
@@ -0,0 +1,1149 @@
...
+
+// GetResourcePluginOptions returns options to be communicated with Resource Manager
+func (p *StaticPolicy) GetResourcePluginOptions(context.Context,
+	*pluginapi.Empty,
+) (*pluginapi.ResourcePluginOptions, error) {
+	return &pluginapi.ResourcePluginOptions{
+		PreStartRequired:      false,
+		WithTopologyAlignment: true,
+		NeedReconcile:         false,
+		AssociatedDevices:     p.resourceNames.List(),
```
JustinChengLZ, line 435: Since this is referring to associated devices, should we name `resourceNames` differently? Maybe `associatedDeviceNames`? Because `resourceNames` can be misunderstood as gpu_memory or gpu_core etc

PR #882, opened 2025-06-30
feat(eviction): add skipped pod kind in numa cpu eviction

lihonghao314:
#### What type of PR is this?
Features

When the Numa Cpu Eviction Plugin performs eviction, it does not evict StatefulSet Pods. This is because the eviction of this Plugin is relatively frequent, which will have a greater impact on stateful services.


pkg/agent/qrm-plugins/cpu/dynamicpolicy/cpueviction/strategy/pressure_numa.go
```diff
@@ -476,6 +479,16 @@ func selectPodRandomly(pods []PodWithUsage) *PodWithUsage {
 	return target
 }
 
+func filterPods(pods []*v1.Pod) []*v1.Pod {
```
xu282934741: In an open source project,this kind of custom logic is best passed in in the form of an option, such as adding a parameter, and a specific kind type pod is not evicted

PR #905, opened 2025-07-17
feat(qrm): numa cpu eviction filter out consumer pod

lihonghao314:


pkg/util/native/pods.go
```diff
@@ -395,3 +397,77 @@ func FilterPodAnnotations(filterKeys []string, pod *v1.Pod) map[string]string {
...
+		for _, ownerReference := range pod.OwnerReferences {
+			if ownerReference.Kind == kind {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// PodIsNotConsumer if pod has env TCE_PRIMARY_PORT, it is not consumer
```
xu282934741, line 412: TCE_PRIMARY_PORT is not valid

pkg/util/native/pods.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/kubewharf/katalyst-core/pkg/consts"
 )
 
+var PrimaryPort = "TCE_PRIMARY_PORT"
```
xu282934741, line 33: TCE_PRIMARY_PORT not valid ; pass the ENV name as a parameter, and modify the functionality to exclude pods containing certain ENVs from eviction.

PR #914, opened 2025-07-27
fix(sysadvisor): fix kcmas resctrl reported memory bandwidth calculation issue

XudongLiuHarold:
#### What type of PR is this?

- fixed the issue to calculate the numa memory bandwidth using malachite-core raw data.
    - calculate the bps per l3 cache
    - aggregate numa bps using l3 cache bps

### DIFF
--- a/pkg/metaserver/agent/metric/provisioner/malachite/provisioner_calculate.go
+++ b/pkg/metaserver/agent/metric/provisioner/malachite/provisioner_calculate.go
@@ -284,44 +284,57 @@ func (m *MalachiteMetricsProvisioner) setContainerRateMetric(podUID, containerNa
 }
 
 // setContainerMbmTotalMetric calcuate the total memory bandwidth usage of a container
-func (m *MalachiteMetricsProvisioner) setContainerMbmTotalMetric(podUID, containerName string, data types.MbmbandData, updateTime *time.Time) {
-	var cpuCodeName string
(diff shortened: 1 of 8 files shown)


pkg/metaserver/agent/metric/provisioner/malachite/provisioner_calculate.go
```diff
@@ -417,3 +430,86 @@ func cpuInList(cpu int, cpuList string) bool {
...
 }
+
+func getCCDMaxMemBandwidth(cpuCode string) uint64 {
+	platform, err := metric.GetCPUPlatform(cpuCode)
+	if err != nil {
+		return 0
+	}
+
+	ccdCount, ok := metric.GetPlatformCCDCount(platform)
+	if !ok || ccdCount == 0 {
```
luomingmeng: Why doesn't the CCD count fallback logic also get integrated within the function?

pkg/metaserver/agent/metric/provisioner/malachite/provisioner_calculate.go
```diff
@@ -417,3 +430,86 @@ func cpuInList(cpu int, cpuList string) bool {
...
+		ccdCount = 1
+	}
+
+	socketBW, ok := metric.GetPlatformSocketBandwidth(platform)
```
luomingmeng: the same as before, the `GetPlatformSocketBandwidth` integrate the fallback logic

PR #930, opened 2025-08-07
feat(resource): add disable reclaim for share pools and improve CPU provision logic

luomingmeng:
#### What type of PR is this?

Features
- Add DisableReclaimSharePools configuration to control reclaim for specific pools
- Refactor share region info extraction to include reclaim enable status
- Improve CPU provision calculation considering reclaim enable status

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/h2non/gock v1.2.0
 	github.com/klauspost/cpuid/v2 v2.2.6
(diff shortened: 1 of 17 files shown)


go.mod
```diff
@@ -173,6 +173,7 @@ require (
 )
 
 replace (
+	github.com/kubewharf/katalyst-api => github.com/luomingmeng/katalyst-api v0.0.0-20250807114248-0f347e060181
```
xu282934741: update go.mod

PR #933, opened 2025-08-08
fix(memory_offloading): improve offloading target size calculation

yadzhang:
#### What type of PR is this?
Add CacheExceptMappedCoeff and minSizeCacheExceptMap constants to better handle cache size calculations. Modify the target size calculation to consider both policy suggestions and a fraction of unmapped cache, ensuring it doesn't exceed policy recommendations or drop below a minimum threshold. This prevents potential performance degradation when swap is disabled.

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/memory/plugin/memory_offloading.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/memory/plugin/memory_offloading.go
@@ -56,6 +56,9 @@ const (
 	InactiveProbe            = 0.1
 	OffloadingSizeScaleCoeff = 1.05
 	CacheMappedCoeff         = 2
+	CacheExceptMappedCoeff   = 2
+
+	minSizeCacheExceptMap = 1 * 1024 * 1024
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/memory/plugin/memory_offloading.go
```diff
@@ -420,20 +423,37 @@ func (tmoEngine *tmoEngineInstance) CalculateOffloadingTargetSize() {
...
 			}
-			err, targetSize := policyFunc(tmoEngine.lastStats, currStats, tmoEngine.conf, tmoEngine.emitter)
+			err, targetFromPolicy := policyFunc(tmoEngine.lastStats, currStats, tmoEngine.conf, tmoEngine.emitter)
 			if err != nil {
 				general.ErrorS(err, "Failed to calculate offloading memory size")
 				return
 			}
 
+			// Calculate the cache size excluding the mapped file size.
+			// This represents the amount of page cache that is not backed by files on disk,
```
cheney-lin: 这个注释不准确吧

PR #985, opened 2025-09-22
feat(logger): implement async logger

JustinChengLZ:
#### What type of PR is this?
We want to optimise async klog's native logging mechanism by making the write function asynchronous.

### DIFF
--- a/cmd/katalyst-agent/app/agent.go
+++ b/cmd/katalyst-agent/app/agent.go
@@ -63,8 +63,8 @@ func Run(
 		return err
 	}
 
-	asyncLogger := logging.NewCustomLogger(genericCtx, conf.CustomLogDir, conf.LogFileMaxSize, conf.LogFileMaxAge, conf.LogFileMaxBackups, conf.LogBufferSize)
-	defer asyncLogger.Shutdown()
+	customLogger := logging.NewCustomLogger(genericCtx, conf.CustomLogDir, conf.LogFileMaxSize, conf.LogFileMaxAge, conf.LogFileMaxBackups, conf.LogBufferSize)
+	defer customLogger.Shutdown()
 
 	for _, genericOption := range genericOptions {
 		genericOption(genericCtx)


pkg/util/logging/logger.go
```diff
@@ -0,0 +1,105 @@
...
+
+const (
+	metricsNameNumDroppedInfoLogs    = "number_of_dropped_info_logs"
+	metricsNameNumDroppedWarningLogs = "number_of_dropped_warning_logs"
+	metricsNameNumDroppedErrorLogs   = "number_of_dropped_error_logs"
+	metricsNameNumDroppedFatalLogs   = "number_of_dropped_fatal_logs"
+)
+
+const (
+	defaultInfoLogFileName    = "agent.info.log"
```
junyu-peng: the name can auto generate using
`fmt.Sprintf("%s.%s.log", filepath.Base(os.Args[0]), severity)`

cmd/base/options/log.go
```diff
@@ -19,29 +19,49 @@ package options
...
+	LogBufferSizeMB     int
+	LogFileMaxAge       int
+	LogFileMaxBackups   int
 }
 
 func NewLogsOptions() *LogsOptions {
 	return &LogsOptions{
 		LogPackageLevel:    general.LoggingPKGFull,
 		LogFileMaxSizeInMB: 1800,
+		LogBufferSizeMB:    10000,
```
junyu-peng: 10000M ~ 10g memory consumption？

JustinChengLZ: After checking, it is not the size of the buffer in MB, but it is the number of log entries that can be written to the ring buffer, and the actual memory consumption depends on the size of each log entry. I will change the parameter name to `LogBufferSize` instead.

cmd/katalyst-agent/app/agent.go
```diff
@@ -60,6 +63,9 @@ func Run(conf *config.Configuration, clientSet *client.GenericClientSet, generic
 		return err
 	}
 
+	asyncLogger := logging.NewCustomLogger(genericCtx, conf.CustomLogDir, conf.LogFileMaxSize, conf.LogFileMaxAge, conf.LogFileMaxBackups, conf.LogBufferSize)
```
junyu-peng: customLogger

pkg/util/logging/logger.go
```diff
@@ -0,0 +1,122 @@
...
+			Filename:   filePath,
+			MaxSize:    maxSizeMB,
+			MaxAge:     maxAge,
+			MaxBackups: maxBackups,
+		}
+
+		// Enable async logger if buffer size is more than 0; otherwise, use synchronous lumberjack logger
+		if bufferSize > 0 {
+			// diodeWriter is a writer that stores logs in a ring buffer and asynchronously flushes them to disk
+			diodeWriter := diode.NewWriter(lumberjackLogger, bufferSize, 10*time.Millisecond, func(missed int) {
```
yehlemias, line 100: why using poller rather than waiter?

JustinChengLZ: Polling performs better with high frequency logs; for waiting, the high frequency of logs will likely cause high frequency of context switching for waking up goroutines at every log, probably leading to high overhead.

PR #989, opened 2025-09-24
skip empty pod list error

junyu-peng:
#### What type of PR is this?

Enhancements

In some scenario, such as DPU offloading, there could be no pods scheduled on node before sale to customs, this PR enables katalyst to report node resources in such cases, instead of returning error.

### DIFF
--- a/pkg/metaserver/agent/pod/pod.go
+++ b/pkg/metaserver/agent/pod/pod.go
@@ -76,10 +76,9 @@ type podFetcherImpl struct {
 	runtimePodFetcher    RuntimePodFetcher
 	kataContainerFetcher *KataContainerFetcher
 
-	kubeletPodsCache                    map[string]*v1.Pod
-	kubeletPodsCacheSkipEmptyError      bool
-	kubeletPodsCacheContinuesEmptyCount int
-	kubeletPodsCacheLock                sync.RWMutex
+	kubeletPodsCache               map[string]*v1.Pod
+	kubeletPodsCacheSkipEmptyError bool
+	kubeletPodsCacheLock           sync.RWMutex
 
 	runtimePodsCache     map[string]*RuntimePod
 	runtimePodsCacheLock sync.RWMutex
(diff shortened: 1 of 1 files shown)


pkg/metaserver/agent/pod/pod.go
```diff
@@ -303,15 +307,16 @@ func (w *podFetcherImpl) syncKubeletPod(ctx context.Context) {
 				"success": "false",
 				"reason":  "empty",
 			})...)
-		return
+		w.kubeletPodsCacheContinuesEmptyCount++
```
JulyWindK: Should it uses a temporary variable copy here? Will there be concurrency competition if I put it in the podFetcherImpl structure?

pkg/metaserver/agent/pod/pod.go
```diff
@@ -76,8 +76,11 @@ type podFetcherImpl struct {
 	runtimePodFetcher    RuntimePodFetcher
 	kataContainerFetcher *KataContainerFetcher
 
-	kubeletPodsCache     map[string]*v1.Pod
-	kubeletPodsCacheLock sync.RWMutex
+	kubeletPodsCache                    map[string]*v1.Pod
+	kubeletPodsCacheSkipEmptyError      bool
+	kubeletPodsCacheContinuesEmptyCount int
```
JulyWindK: Where is this value assigned

pkg/metaserver/agent/pod/pod.go
```diff
@@ -320,6 +325,9 @@ func (w *podFetcherImpl) syncKubeletPod(ctx context.Context) {
 
 	w.kubeletPodsCacheLock.Lock()
 	w.kubeletPodsCache = kubeletPodsCache
+	if w.kubeletPodsCacheContinuesEmptyCount >= w.podConf.KubeletPodCacheSyncEmptyThreshold && !w.kubeletPodsCacheSkipEmptyError {
+		w.kubeletPodsCacheSkipEmptyError = true
```
JulyWindK: Is this kubeletPodsCacheSkipEmptyError variable switch not configured by the administrator

pkg/metaserver/agent/pod/pod.go
junyu-peng: No concurrency issue here, because it's only read/write in this single goroutine. And yes, make it a local variable will be more reasonable, i'll change it.

pkg/metaserver/agent/pod/pod.go
junyu-peng: in syncKubeletPod

pkg/metaserver/agent/pod/pod.go
junyu-peng: The admin can config `KubeletPodCacheSyncEmptyThreshold`, it will affect the value of `kubeletPodsCacheSkipEmptyError`.

JulyWindK (approved):

xu282934741 (approved):

PR #1034, opened 2025-11-28
WIP: feat(advisor):  apply cpu headroom discount when power level is present 

h-w-chen:
#### What type of PR is this?
Features

This PR allows power-ware plugin to enable reclaimed cpu headroom discount when the node's power level is of p1/p2/p3.

To ensure the power-aware plugin enabling to happen prior qos-aware cpu resource advisor being called to GetHeadroom, __Init phase__ is put with cpu sub resource advisor up to qos-aware plugin.

### DIFF
--- a/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/decorator/decorator_discount.go
+++ b/pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/decorator/decorator_discount.go
@@ -18,6 +18,7 @@ package decorator
 
 import (
 	"context"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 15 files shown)


pkg/agent/sysadvisor/plugin/qosaware/resource/cpu/assembler/headroomassembler/decorator/decorator_discount.go
```diff
@@ -109,11 +111,26 @@ func (d *discountDecorator) GetHeadroom() (resource.Quantity, map[int]resource.Q
 	}
 
 	headroom, numaHeadrooms, err := d.inner.GetHeadroom()
-	if err == nil && currentDiscount < 1.0 {
-		headroom = applyDiscount(headroom, currentDiscount)
-		for numa := range numaHeadrooms {
-			numaHeadrooms[numa] = applyDiscount(numaHeadrooms[numa], currentDiscount)
-		}
+	if err != nil || currentDiscount >= 1.0 {
+		return headroom, numaHeadrooms, err
```
JulyWindK, line 121: Does this error need to be encapsulated? When err is empty and currentDiscount>=1.0, is it expected that err will be empty?

h-w-chen: intention here is no discount and just pass on what the inner returned headroom values. The mixture of error/value check was confusing - thanks for raising this issue. Code rewritten to make the intention explicit.

PR #1122, opened 2026-04-07
refactor(irq-tuning): handle known failure of get GetNicQueue2Irq

zhanghaoyu1986:
handle known failure of get GetNicQueue2Irq

#### What type of PR is this?
Bug fixes

fix nic queue<->irq mapping

### DIFF
--- a/pkg/util/machine/network_linux.go
+++ b/pkg/util/machine/network_linux.go
@@ -94,6 +94,11 @@ var unsupportedQueue2IrqDrivers = []string{
 	"xsc",
 }
 
+const (
+	gveNicQueueFilter    = "gve-ntfy-blk"
+	gveNicQueueDelimeter = "@"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/util/machine/network_linux.go
```diff
@@ -761,6 +783,10 @@ func GetNicQueue2Irq(nicInfo *NicBasicInfo) (map[int]int, map[int]int, error) {
 		}
 	}
 
+	if isUnsupportedNicQueue2Irq(nicInfo) {
```
JulyWindK, line 873: It is recommended to place this logic before the function to terminate meaningless execution early.

zhanghaoyu1986: We intentionally put isUnsupportedNicQueue2Irq at the end of the function. This design guarantees that if NICs in unsupportedQueue2IrqDrivers ever support obtaining queue2irq in the future, they can acquire it properly.

luomingmeng (approved):

PR #1149, opened 2026-05-09
chore: watch events of cgroup file CREATE and REMOVE

cheney-lin:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/qrm-plugins/memory/handlers/sockmem/sockmem_linux.go
+++ b/pkg/agent/qrm-plugins/memory/handlers/sockmem/sockmem_linux.go
@@ -199,6 +199,10 @@ func SetSockMemLimit(conf *coreconfig.Configuration,
 			general.Errorf("get nil pod from metaServer")
 			continue
 		}
+		if native.PodIsTerminated(pod) {
+			general.Infof("pod %s/%s is marked for deletion", pod.Namespace, pod.Name)
+			continue
+		}
 		for _, containerStatus := range pod.Status.ContainerStatuses {
 			if !containerStatus.Ready {
 				continue
--- a/pkg/metaserver/agent/pod/pod.go
+++ b/pkg/metaserver/agent/pod/pod.go
@@ -19,12 +19,16 @@ package pod
 import (
 	"context"
 	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
 	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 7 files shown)


JulyWindK (dismissed):

pkg/metaserver/agent/pod/pod.go
```diff
@@ -154,35 +158,126 @@ func (w *podFetcherImpl) GetContainerID(podUID, containerName string) (string, e
 	return native.GetContainerID(pod, containerName)
 }
 
-func (w *podFetcherImpl) Run(ctx context.Context) {
-	watcherInfo := general.FileWatcherInfo{
-		Path:     w.cgroupRootPaths,
-		Filename: "",
-		Op:       fsnotify.Create,
```
luomingmeng, line 161: Why not directly implement fsnotify.Remove inside Op for this feature

pkg/agent/qrm-plugins/memory/handlers/sockmem/sockmem_linux.go
```diff
@@ -199,6 +199,10 @@ func SetSockMemLimit(conf *coreconfig.Configuration,
 			general.Errorf("get nil pod from metaServer")
 			continue
 		}
+		if pod.DeletionTimestamp != nil {
```
luomingmeng: Why not use PodIsTerminated ?

pkg/metaserver/agent/pod/pod.go
```diff
@@ -154,35 +158,126 @@ func (w *podFetcherImpl) GetContainerID(podUID, containerName string) (string, e
...
+					(needSync || rootCgroupPaths.Has(eventPath) || podCgroupPaths.Has(eventPath)) {
+					klog.InfoS("pod cgroup path event", "event", event.String())
+					if rateLimiter.Allow() {
+						w.syncKubeletPod(ctx)
+						timer.Reset(w.podConf.KubeletPodCacheSyncPeriod)
+					}
+				}
+			case err, ok := <-podCgroupWatcher.Errors:
+				if !ok {
+					return
```
luomingmeng: Why not implement this logic by extending the RegisterFileEventWatcher utility function.

cheney-lin: 现在逻辑比RegisterFileEventWatcher 复杂的多，我认为不适合改原util函数

luomingmeng: Or implement it in a separate util function to decouple it from the logic of synchronizing kubelet pods.

cheney-lin: 哪部分逻辑要封装

luomingmeng: implement a `RegisterSubDirEventWatcher` function like `RegisterFileEventWatcher` in pkg/util/general/file.go
```
// SubDirWatcherInfo describes a directory tree (root + first-level children) to watch.
// It is used by RegisterSubDirEventWatcher to monitor pod-cgroup-like layouts where:
//   - a root path is watched for child directory creation/removal,
//   - first-level child directories under each root are also watched and
//     dynamically added/removed as the tree changes.
type SubDirWatcherInfo struct {
    // RootPaths are the parent directories whose first-level subdirectories
    // should be discovered and watched. Each root itself is also watched so
    // that newly created / removed subdirectories can be tracked.
    RootPaths []string
}

// DirWatchListFunc returns a snapshot of the paths currently being watched
// (both root paths and discovered first-level children). It is goroutine
// safe and may be called from any goroutine while the watcher is alive.
// After the watcher is stopped, it returns an empty slice.
type DirWatchListFunc func() []string

// RegisterSubDirEventWatcher watches a set of root directories together with
// their first-level child directories, and reports a "needs-sync" signal to
// the caller through the returned channel whenever a create / remove event
// happens on a watched root or one of its child directories.
//
// Behavior:
//   - On startup every root in watcherInfo.RootPaths is added to the watcher and
//     its first-level child directories are discovered via filepath.WalkDir
//     and also added.
//   - At runtime, when a Create event fires on a path whose parent is a
//     watched root and the new path is a directory, the new child is added
//     to the watcher set.
//   - When a Remove event fires on a previously watched child, the child is
//     removed from the watcher set.
//   - Any Create / Remove event hitting a watched root or child triggers a
//     send on the returned channel. Sends are non-blocking: if the receiver
//     is slow, events may be coalesced (the channel is buffered with size 1).
//   - On <-stop, the underlying fsnotify watcher is closed and the returned
//     channel is closed.
//
// The returned DirWatchListFunc gives the caller an observability hook to dump
// the currently watched paths (mirroring fsnotify.Watcher.WatchList()).
//
// The caller is responsible for any rate limiting / periodic fallback /
// actual sync behavior on top of the returned signal channel.
func RegisterSubDirEventWatcher(stop <-chan struct{}, watcherInfo SubDirWatcherInfo) (<-chan struct{}, DirWatchListFunc, error)
```

PR #1173, opened 2026-06-01
fix(gpu): fix reporter to trigger periodically and filter out inactive pods

JustinChengLZ:
#### What type of PR is this?

Bug fixes

### DIFF
--- a/pkg/agent/qrm-plugins/gpu/baseplugin/reporter/reporter.go
+++ b/pkg/agent/qrm-plugins/gpu/baseplugin/reporter/reporter.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/pkg/kubelet/checkpointmanager"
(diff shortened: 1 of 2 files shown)


pkg/agent/qrm-plugins/gpu/baseplugin/reporter/reporter.go
yehlemias: @luomingmeng @JustinChengLZ i think we can move this logic out of the for-loop, there is no need to force sync pod list for every pod and every numa

yehlemias, line 595: i mean move out of the whole loop

yehlemias: just force sync the pod list cache once

luomingmeng (dismissed):

pkg/agent/qrm-plugins/gpu/baseplugin/reporter/reporter.go
```diff
@@ -568,9 +574,20 @@ func (p *gpuReporterPlugin) addKubeletCheckpointAllocations(idToAllocations map[
...
 	podDeviceEntries, _ := checkpointData.GetDataInLatestFormat()
+	if len(podDeviceEntries) == 0 {
+		return nil
+	}
 
 	validDeviceNames := sets.NewString(p.gpuDeviceNames...)
 
+	activePods, err := p.metaServer.GetPodList(context.WithValue(p.ctx, metaserverpod.BypassCacheKey, metaserverpod.BypassCacheTrue), native.PodIsActive)
+	if err != nil {
+		general.Warningf("failed to get active pod list: %v", err)
```
xu282934741, line 586: Why didn't we return the error here?

JustinChengLZ: Returning the error here would mean that we are not able to show the pods in the kubelet checkpoint, which may cause pods to be admitted unexpectedly and have TopologyAffinityError issues. However, if metaserver returns an error and we ignore it, it will at most cause pods to be pending, and if the error is transient / resolves itself, the pods can then be admitted into the node without issue.

luomingmeng (approved):

